PR #4, opened 2017-11-28
Use $Shape<T> instead of optional fields

MichaelDeBoey:
Just like in erikras/final-form#2 🙂


MichaelDeBoey:
@erikras running `npm run precommit` didn't do anything now 😕

PR #11, opened 2017-11-28
Be more explicit about HoC and Render Prop comparison

jaredpalmer:


erikras:
You prefer a ✅ over "optional"? Okay. I don't mind.

PR #12, opened 2017-11-29
Bring API docs in line with real API

MichaelDeBoey:
In addition to erikras/final-form#8

Also made some small changes to the `flow` typings to be in line also.


circlingthesun:
In the FormRenderProps 'validating' (as seen here: https://codesandbox.io/s/wy7z7q5zx5) is also undocumented.

MichaelDeBoey:
@circlingthesun I only added the changed I made for erikras/final-form#8, but will check the flow types and update the API docs accordingly 🙂

MichaelDeBoey:
> This object contains everything in [🏁 Final Form's `FormState`](https://github.com/erikras/final-form#formstate) as well as:

@circlingthesun The docs for `FormRenderProps` are clear imo, but the `flow` type needs to be adjusted 🙂

MichaelDeBoey:
@erikras It's best to run `prettier` once again I think 🙂

PR #30, opened 2017-12-04
Allow later initialization through initialValues prop

### LINKED ISSUE #18: Change values
Hello! Is there any possibility to pass initial values after form created. Or change particular value from upper component. 
I prefill my form using query after form is rendered and I can not pass loaded values to inputs.
Thanks

erikras:
Fixes #18.


erikras:
Published in [`v1.1.1`](https://github.com/final-form/react-final-form/releases/tag/v1.1.1).

PR #32, opened 2017-12-04
Returned a promise from handleSubmit when submission is async

### LINKED ISSUE #21: Reset form after form submission succeed
Hi there ! 
I was wondering how i can reset my form on successful submit ? 
Seems that we don't have access to the `reset()` function in `onSubmit()`. 
How can i handle that ? 
Thanks for the work by the way , react-final-form is awesome  :)

erikras:
Fixes #21.


erikras:
Published in [`v1.1.1`](https://github.com/final-form/react-final-form/releases/tag/v1.1.1).

PR #33, opened 2017-12-04
Fix links to package

MichaelDeBoey:
Just like in final-form/final-form#17 🙂


erikras:
These are _so_ important, as GH will forward, but thanks! 👍

MichaelDeBoey:
Didn't know @GitHub would forward, but always nice to have a more correct link I think 🙂

PR #36, opened 2017-12-05
Stopped meta.name prop. Name is only on input

### LINKED ISSUE #34: Field has name prop in meta props

Probably a bug

Field has `name` prop both in `input` and `meta` props.

From the docs, `meta` should not have duplicate `name` prop, it exists only in `input` props.

### Sandbox Link

https://codesandbox.io/s/5z7z467o4l

Open the console and you'll see it.

erikras:
Fixes #34.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -109,7 +109,14 @@ export default class Field extends React.PureComponent<Props, State> {
       value: _value,
       ...rest
     } = this.props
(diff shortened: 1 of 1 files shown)


erikras:
Published in [`v1.2.0`](https://github.com/final-form/react-final-form/releases/tag/v1.2.0).

PR #37, opened 2017-12-05
Added warning on improper value prop usage

erikras:
Addresses confusion of devs wanting to set a field value with the `value` prop.


erikras:
Published in [`v1.2.0`](https://github.com/final-form/react-final-form/releases/tag/v1.2.0).

PR #44, opened 2017-12-06
Revert "Revert "Added onChange prop to FormSpy""

### LINKED ISSUE #41: Expose method of subscribing to form data
Feature request: I would like to be able to subscribe to form updates for the purpose of autosave without submission. I am currently using `FormSpy` to do this but its render function is also triggered by parents rerendering:

```javascript
<FormSpy subscription={{ values: true }}>
   {({ values }) => {
     onChange(values);
     return null;
   }}
</FormSpy>
```
It would nice if there was easy way to listen to updates without it also trigging on renders and gives the granularity subscriptions offer (`debug` is too noisy). I could get hold of `reactFinalForm` from `context` but this in not part of the public API.

One of the following would work for me:

* An `onChange` prop on `FormSpy` 
* A simular component that accepts `subcription` and `onChange` as `props`
* A component that fetches `reactFinalForm` from the `context` 
* `reactFinalForm` being exposed as a render prop on Form

_Am I missing something, is there already a way to do this?_

erikras:
Reverts #43. Changed my mind again. 🙈

Closes #41.


philholden:
Thanks for doing so quickly, any chance of a release? My hacky use of the render function is breaking down now I need to `setState()`. Cannot initiate setState from render (infinite loop).

erikras:
Yep! [`v1.2.0`](https://github.com/final-form/react-final-form/releases/tag/v1.2.0).

philholden:
Thanks.

Issue #45, opened 2017-12-08
Add support to value lifecycle


gsantiago:
The [value lifecycle](https://redux-form.com/7.2.0/docs/valuelifecycle.md/) is one of the best features from redux-form. Will react-final-form support something like that?

erikras:
Hmm... I was hoping to not need to implement that, instead laying the burden of formatting and parsing on the code that initializes and submits the form. Can you make an argument for why you think you need it?

dispix:
I agree with @gsantiago, the value lifecycle was a really simple and nice way to separate the data from the display. I think that there are quite a lot of cases where the value stored in the form state and the value passed to the "dumb" component need to be handled differently.

erikras:
Published fix in [`v2.1.0`](https://github.com/final-form/react-final-form/releases/tag/v2.1.0).

PR #72, opened 2017-12-18
Value lifecycle

dummerbd:
The "value lifecycle" from Redux Form was a must-have feature for me and others (see #45 ). This PR introduces three new props for the `Field` component:
 - `format: (value: ?any, name: string) => ?any | null`
 - `parse: (value: ?any, name: string) => ?any | null`
 - `normalize: (value: ?any, previousValue: ?any, allValues: object) => ?any`

They work basically the same way as they do in Redux Form. The `allowNull` prop is still respected and applied after `format`. By default, `format` converts undefined to an empty string and `parse` converts an empty string to undefined. Either can be set to `null` explicitly opt-out of that behavior and of course a custom function can be provided for more advanced cases.

The `normalize` function is almost the same except it is not passed the `previousAllValues` since that is not easily available from within the `Field` component. I personally don't use the `normalize` prop so I can't think of a use case where the `previousAllValues` is needed.

So, there's no breaking changes and minimal amount of code/complexity introduced.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -25,6 +25,16 @@ export default class Field extends React.PureComponent<Props, State> {
   state: State
   unsubscribe: () => void
 
+  static contextTypes = {
+    reactFinalForm: PropTypes.object
+  }
+
+  static defaultProps = {
+    format: (value: ?any, name: string) => (value === undefined ? '' : value),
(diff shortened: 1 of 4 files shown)


dummerbd:
Hmmm, not sure why it's still saying there's a merge conflict. Should I rebase and open a new PR?

Issue #83, opened 2017-12-27
Submitting a form with no event results in a runtime error (React Native)


cosmith:
### Are you submitting a **bug report** or a **feature request**?

Feature request

### What is the current behavior?

Submitting a form with no event results in a runtime error.

### What is the expected behavior?

I'm using React Final Form on React Native and I don't always have an event passed to the submit handler. 
```js
          <Form
                onSubmit={this.onSubmit}
                initialValues={/* ... */}
                render={({handleSubmit, reset, submitting, pristine, values}) => (
                    <View style={styles.container}>
                        <NavigationBar
                            leftButton="back"
                            onLeftButtonPress={this.props.navigator.pop}
                            rightButtonText={t("Save")}
                            onRightButtonPress={handleSubmit}
                        />
                ...
          </Form>
```

### Sandbox Link

Obviously bad example just to show the error: https://codesandbox.io/s/k10xq99zmr

### What's your environment?

React Native 0.48
Final Form 3.0.0
React Final Form 2.1.0 

### Other information

It's easy to bypass this error with this:
```js
onRightButtonPress={() => {
    handleSubmit({preventDefault: () => {}});
}}
```
But I would rather have a check in the `handleSubmit` function to bypass the `preventDefault` call if no event is passed.

cosmith:
I'm happy to submit a PR to https://github.com/final-form/react-final-form/blob/ec271773fd0f7b0c7f7fe0c4992bfcf021b63dc8/src/ReactFinalForm.js#L109
```diff
- event.preventDefault()
+ event && event.preventDefault()
```
I just need to check with Flow that everything still works

erikras:
> I'm happy to submit a PR

![web image 2017-12-27 15-09-22](https://user-images.githubusercontent.com/4396759/34383765-ee638118-eb17-11e7-802d-0e77dff8ea48.jpg)

maciejmyslinski:
Shouldn't this issue be closed since #84 is merged?

erikras:
Indeed. Thanks for your vigilance.

PR #84, opened 2017-12-27
Make event parameter optional on handleSubmit

cosmith:
See discussion here : https://github.com/final-form/react-final-form/issues/83

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -105,8 +105,8 @@ export default class ReactFinalForm extends React.PureComponent<Props, State> {
 
   notify = (state: FormState) => this.setState({ state })
 
-  handleSubmit = (event: SyntheticEvent<HTMLFormElement>) => {
-    event.preventDefault()
+  handleSubmit = (event?: SyntheticEvent<HTMLFormElement>) => {
+    event && event.preventDefault()
     return this.form.submit()
   }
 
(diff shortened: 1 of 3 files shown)


src/ReactFinalForm.js
```diff
@@ -105,8 +105,8 @@ export default class ReactFinalForm extends React.PureComponent<Props, State> {
 
   notify = (state: FormState) => this.setState({ state })
 
-  handleSubmit = (event: SyntheticEvent<HTMLFormElement>) => {
-    event.preventDefault()
+  handleSubmit = (event?: SyntheticEvent<HTMLFormElement>) => {
```
cosmith: Not sure if it should be `event?: SyntheticEvent` or `event: ?SyntheticEvent`

erikras: In this case, I think it does not matter, but I would choose the latter.

erikras:
Can you add a test that fails without your change, but passes with it?

src/ReactFinalForm.js
cosmith: I think since the parameter is optional it's good this way: https://flow.org/en/docs/types/functions/#optional-parameters-a-classtoc-idtoc-optional-parameters-hreftoc-optional-parametersa

erikras: Okay. I stand corrected with citation! 👍

erikras:
Nice work! 👍

cosmith:
Thanks!

erikras:

PR #87, opened 2017-12-29
Add 'batch' and 'mutators' TS definitions

circlingthesun:
Added 'batch' and 'mutators' definitions to FormRenderProps


erikras:

PR #95, opened 2018-01-03
Make event optional in TS handleSubmit definition

danizord:
Reflecting the signature of JS implementation.


erikras:
Thanks. This was overlooked in #84.

erikras:

PR #104, opened 2018-01-10
Enabled checkboxes with value props to maintain array values

### LINKED ISSUE #90: input[type=checkbox] should behave as select[multiple]

I don't know, problably a bug... or a feature :)

When using using input type="checkbox" we only can get a boolean.

https://codesandbox.io/s/n4ppqq63n0

In this example, I really want it to behave as the select with multiple option.

### Other information

Appart from this, lib is perfect :+1: keep the good work

erikras:
Fixes #90.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -147,11 +147,18 @@ export default class Field extends React.PureComponent<Props, State> {
     }
     const input = { name, value, ...this.handlers }
     warning(
-      !_value || (rest.type === 'radio' && component === 'input'),
-      'The value prop on Field is ONLY for use with component="input" and type="radio".'
+      !_value ||
+        ((rest.type === 'radio' || rest.type === 'checkbox') &&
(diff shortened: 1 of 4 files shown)


erikras:

developpinpray:
@erikras  thanks for the new feature. I have tested it and it works well. Could we possibly remove the check for  component === 'input' ( line 152 of Field.js) ? I am using an adapter so I can use MaterialUI checkboxes, everything works fine, just get that warning a lot.

mohadib:
Thanks for the new feature @erikras. Would it be possible to remove the check 'component === 'input''. I am using an adapter w/ MaterialUI checkbox. Seems to work just fine, but I see this warning in the console a lot.

'The value prop on Field is ONLY for use with component="input" and type="radio" or type="checkbox".'

Thanks

PR #105, opened 2018-01-10
Provided Form API to FormSpy

### LINKED ISSUE #101: [request] FormSpy should receive FormApi
feature request

`FormSpy` provides access to internal `FormState`, but does not expose a way to comprehend the registered fields.

`FormSpy` is passed `FormApi` in addition to `FormState` so that one may call `getRegisteredFields: () => string[]`

### What's your environment?
RFF v2.1.1
React v16.2.0

### Workaround
It looks like I can get `FormApi` from the context as `reactFinalForm`, but I think it a useful addition to provide the `FormApi` to the `FormSpy` unless there is a reason to protect it.
  

BTW - so far everything has been fantastic.  Very solid library, almost no issues converting a mature codebase from react-formal over to this.

erikras:
Fixes #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 ✅ Opt-in subscriptions - only update on the state you need!
 
-✅ 💥 **2.7k gzipped** 💥
+✅ 💥 **2.8k gzipped** 💥
 
 ---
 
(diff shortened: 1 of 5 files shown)


erikras:

PR #122, opened 2018-01-17
Make errors `any` not `boolean`

ForbesLindesay:
It's unlikely that the errors will ever be a boolean. Typically they are `null | string` but final-form supports `any` value.


erikras:
Hahaha! Quite a mistake. Dates back to #4.

erikras:

PR #123, opened 2018-01-17
Remove duplicated methods from FormRenderProps

ForbesLindesay:
These methods already appear in SubsetFormApi


erikras:

PR #124, opened 2018-01-17
RenderableProps component prop should still be given T

ForbesLindesay:


erikras:

PR #144, opened 2018-02-07
feature/persistentSubmitErrors added persistentSubmitErrors prop

alex-shatalov:
final-form/final-form#77

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -62,7 +62,8 @@ export default class ReactFinalForm extends React.PureComponent<Props, State> {
       render,
       subscription,
       validate,
-      validateOnBlur
+      validateOnBlur,
+      persistentSubmitErrors
     } = props
     warning(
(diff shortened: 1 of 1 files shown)


erikras:
What does this accomplish? Can you give more explanation?

erikras:
Sorry, I had not seen https://github.com/final-form/final-form/pull/77/. You should have linked to it for clarity.

alex-shatalov:
We have case that we need to submit form even if it has validation errors. So when we submitting form it fails on hasSyncErrors() in submit.  In redux form prop persistentSubmitErrors helps us to avoid this problem. 

Maybe you could give advise how to get it without this changes in pull request. Thanks.

erikras:
> Maybe you could give advise how to get it without this changes in pull request. Thanks.

Sounds like you're looking for [Field Warnings](https://github.com/final-form/react-final-form#field-warnings).

PR #153, opened 2018-02-12
Stopped extending React.PureComponent

### LINKED ISSUE #150: Should the Components really extend PureComponent?
Hey, really awesome library, thanks so much for sharing!

# tl;dr

I think that all Components that use render props should extend `Component`, instead of `PureComponent`.

# Longer version

My colleague noticed that the `<Form>` component is a `PureComponent`, so he thought it'd be a good idea to maintain referential identity between renders, and define the render function in such a way that it would allow `<Form>` to opt-out of rerendering. Like so:

```js
class Form2 extends Component {
  renderForm = ({ handleSubmit }) => {
    ...fancy rendering stuff
  };

  render() {
    return <Form render={this.renderForm} />;
  }
}
```

This works fine, as long as `renderForm` is a pure function, solely depending on it's own parameters. However, as soon as you mix in state or props from the wrapping component (`Form2` in the example), the `renderForm` function passed to `<Form>` will become stale.

This is all much better explained in the excellent article by @ryanflorence: https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578

I think the quote most relevant for this issue is the following:
> That means an inline render prop function won’t cause problems with `shouldComponentUpdate`: It can’t ever know enough to be a `PureComponent`.

As I understand it, if you use render props, you should never extend `PureComponent`, since you don't know what's going on inside the render function.

What are your thoughts?

erikras:
Fixes #150.

It did require changing some tests because some of the components rendered more often than before, but this is still the right move for all the reasons described in #150.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -19,7 +19,7 @@ type State = {
   state: FieldState
 }
 
-export default class Field extends React.PureComponent<Props, State> {
+export default class Field extends React.Component<Props, State> {
   context: ReactContext
   props: Props
   state: State
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -39,14 +39,14 @@ describe('Field', () => {
     expect(renderInput).not.toHaveBeenCalled()
     const dom = TestUtils.renderIntoDocument(<Container />)
     expect(renderInput).toHaveBeenCalled()
-    expect(renderInput).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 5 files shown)


erikras:
Published in [`v3.1.1`](https://github.com/final-form/react-final-form/releases/tag/v3.1.1).

PR #154, opened 2018-02-12
Update comparison to Formik

jaredpalmer:
v0.11 ships FieldArray


erikras:
Hehe... when I saw that, I was on vacation, and wondered which of us would get to it first. 👍

PR #158, opened 2018-02-13
Ensured that input.checked is boolean

erikras:
There were times when it could be `-1` or `0` instead of `true` or `false`.


erikras:
Published in [`v3.1.1`](https://github.com/final-form/react-final-form/releases/tag/v3.1.1).

PR #164, opened 2018-02-21
Use jsx block for better code highlighting

diegohaz:
Before:
image

After:
image


erikras:
Ha! I almost always use that. Thanks!

PR #183, opened 2018-03-08
Handle props changes

### LINKED ISSUE #128: [BUG] onSubmit/validate/validateOnBlur use out of date values after react reconsilation

If you replace the `onSubmit` function after the initial render, the new function will be ignored because react-final-form takes that configuration on startup.  Here is an example that makes the issue fairly clear.

```js
class App extends React.Component {
  state = { mounted: false };
  componentDidMount() {
    this.setState({ mounted: true });
  }
  render() {
    return (
        <Form
          onSubmit={
            this.state.mounted ? (
              () =>
                window.alert("This is the on submit that should fire")
            ) : (
              () =>
                window.alert("This is the on submit that actually fires")
            )
          }
          render={({ handleSubmit, reset, submitting, pristine, values }) => (
              <button type="submit" disabled={submitting || pristine}>
                Submit
              </button>
            </form>
          )}
        />
    );
  }
}
```

It can be quite a bit more subtle though.  The following has the same issue because the original arrow function has captured an old version of the state:

```js
class App extends React.Component {
  state = { mounted: false };
  componentDidMount() {
    this.setState({ mounted: true });
  }
  render() {
    const mounted = this.state.mounted;
    return (
        <Form
          onSubmit={
            () => (mounted
               ? window.alert("This is the on submit that should fire")
               : window.alert("This is the on submit that actually fires")
            )
          }
          render={({ handleSubmit, reset, submitting, pristine, values }) => (
              <button type="submit" disabled={submitting || pristine}>
                Submit
              </button>
            </form>
          )}
        />
    );
  }
}
```

Another more subtle way that this can occur is if you replace the entire form with another form - if this is done in place, react will think it can update the existing form, rather than creating a new one.

### Sandbox Link

[Edit 🏁 React Final Form - Simple Example](https://codesandbox.io/s/v6z0m2o767)

### Suggested fixes

There are two ways this could be resolved:

1. Add methods to final-form so that the onSubmit/validate/validateOnBlur config options can be updated.  This update can then happen in `componentWillReceiveProps` alongside the existing update to initial state.  This needs to remain a fairly "cheap" operation, since arrow functions would trigger this on every render.
2. "Proxy" the methods by defining methods that look like `_onSubmit = data => this.props.onSubmit(data)` on the `ReactForm` instance and then use those in place of directly referencing `this.props.onSubmit`.  That way you would always get an up-to-date reference.  The issue with this is that final-form does different things depending on the `length` of `onSubmit` and that can't be replicated here.

Although this issue *is* avoidable, it is going to lead to some really subtle bugs that are hard to track down, so I think it should be fixed.

ForbesLindesay:
[closes #128]

The peer dependency on final-form will need to be updated to the version that includes the `setConfig` method.

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -136,21 +136,16 @@ export default class ReactFinalForm extends React.Component<Props, State> {
     ) {
       this.form.initialize(nextProps.initialValues)
     }
-    if (this.props.debug !== nextProps.debug) {
-      this.form.setConfig('debug', nextProps.debug)
-    }
-    if (!shallowEqual(this.props.mutators, nextProps.mutators)) {
-      this.form.setConfig('mutators', nextProps.mutators)
-    }
-    if (this.props.onSubmit !== nextProps.onSubmit) {
-      this.form.setConfig('onSubmit', nextProps.onSubmit)
-    }
-    if (this.props.validate !== nextProps.validate) {
-      this.form.setConfig('validate', nextProps.validate)
-    }
-    if (this.props.validateOnBlur !== nextProps.validateOnBlur) {
-      this.form.setConfig('validateOnBlur', nextProps.validateOnBlur)
-    }
+    ;['debug', 'mutators', 'onSubmit', 'validate', 'validateOnBlur'].forEach(
+      key => {
+        if (this.props[key] === nextProps[key]) {
+          return
+        }
+        if (this.form.setConfig) {
+          this.form.setConfig(key, nextProps[key])
+        }
+      }
+    )
     if (process.env.NODE_ENV !== 'production') {
       warning(
         shallowEqual(this.props.decorators, nextProps.decorators),
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -136,6 +136,31 @@ export default class ReactFinalForm extends React.Component<Props, State> {
     ) {
       this.form.initialize(nextProps.initialValues)
     }
+    if (this.props.debug !== nextProps.debug) {
+      this.form.setConfig('debug', nextProps.debug)
+    }
+    if (!shallowEqual(this.props.mutators, nextProps.mutators)) {
+      this.form.setConfig('mutators', nextProps.mutators)
+    }
+    if (this.props.onSubmit !== nextProps.onSubmit) {
+      this.form.setConfig('onSubmit', nextProps.onSubmit)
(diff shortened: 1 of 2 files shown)


ForbesLindesay:
@erikras I assume you're going to require tests for this. Are you otherwise happy with the approach?

erikras:
I guess so, sure.

erikras:
Is this going anywhere? What I dislike about this PR is how "unminimizable" the code is. I question if the utility / bundle size ratio is high enough.

ForbesLindesay:
Sorry, I've been really busy. I'd be happy for someone to finish it off. I view this functionality as pretty critical as I've had multiple bugs caused by final-form not handling this case properly. It doesn't seem like much code to me.

Andarist:
could me shortened with smth like:
```js
updateConfig = (prev, current) => ['debug', 'mutators', 'onSubmit', 'validate', 'validateOnBlur'].forEach(key => {
	if (prev[key] === current[key]) {
		return
	}

	this.form.setConfig(key, current[key])
})
```

Impact of both versions on the gzipped production bundle would have to be measured though, gzip is a tricky thing.

From what I understand this would require bumping peerDep which would be a breaking change of sorts, so maybe some alternative approach could be used to solve this?

ForbesLindesay:
Changes:

1. updated based on @Andarist's suggestion to use an array for the keys.
2. ignore updates when `setConfig` does not exist, allowing the peer dependency to remain as it is and for this to be a non-breaking change.
3. added a test case. I've only tested `onSubmit` changes since the others are pretty much the same code path.

@erikras let me know if anything else is needed to get this landed, however I'm on holiday for a while after the end of this week, so if it needs substantial changes it may take a while for me to get to them.

erikras:
Looks great to me!

erikras:
Actually, there was a bug in your test. You were in initializing `state` to `newOnSubmit` instead of `oldOnSubmit`. 😄

Andarist:
@erikras it would be good to test gzipped size with previous version of this code

erikras:
@Andarist Nice! I originally had [the subscription filtering stuff](https://github.com/final-form/final-form/blob/master/src/subscriptionFilter.js) written out as 20 `if` blocks, because I couldn't figure out how to keep Flow happy, but when I moved it to iterating over an array of keys, I remember it dropping the bundle size substantially.

erikras:

PR #185, opened 2018-03-13
Fix exported flow types

### LINKED ISSUE #142: Flow type definitions aren't working
Bug report

Flow type definitions aren't working

Flow type definitions should work

### What's your environment?

**Node**: v9.5.0
**Flow**: v0.64.0
**React**: v16.2.0
**Final-Form**: v4.2.0
**React-Final-Form**: v3.1.0

### Steps to reproduce
1. `npm install flow-bin final-form react-final-form`
2. Write this code:

```js
/**
 * @flow
 */

import { version } from 'react-final-form'

const v: number = version // This should throw error, but doesn't

console.log(v)
```

3. Run `flow` in terminal or editor

As you can see, it doesn't cause any issues:

screen shot 2018-02-06 at 5 01 10 pm

### Possible solution

I think this may have to do with the fact that in the `dist`, the `main` file is `react-final-form.cjs.js`. I think this may mean that the `.flow` file that gets generated may need to be called `react-final-form.cjs.js.flow`. When I renamed `index.js.flow` to `react-final-form.cjs.js`, flow seems to work:

screen shot 2018-02-06 at 5 01 00 pm

---

**NOTE**: The types still seem to be problematic however. When I use the [simple example](https://codesandbox.io/s/ww40y2m595) (only thing I changed was `<Styles>` to `<div>`), there seem to be all sorts of Flow errors. I'm not sure if this has to do with the type definitions being incorrect or not.

Example Code: [Gist](https://gist.github.com/anonymous/f434a5ba2b9b73200dd3f4cfd9a145b5)
Terminal Output: [Gist](https://gist.github.com/anonymous/7e3e523fa6166525f56b4364ac1cd854)

erikras:
Fixes #142.
Depends on https://github.com/final-form/final-form/pull/89


erikras:
Published in [`v3.1.3`](https://github.com/final-form/react-final-form/releases/tag/v3.1.3).

PR #200, opened 2018-03-22
Export more flow types

erikras:
Addresses issues raised in #142.


erikras:
Published in [`v3.1.5`](https://github.com/final-form/react-final-form/releases/tag/v3.1.5).

PR #204, opened 2018-03-27
Add dev warning when type prop is missed for checkbox, radio or multi…

Andarist:
Simple PR improving DX by providing a helpful warning to the developer about his/her mistake.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -100,6 +100,30 @@ export default class Field extends React.Component<Props, State> {
     },
     onChange: (event: SyntheticInputEvent<*> | any) => {
       const { parse, value: _value } = this.props
+
+      if (process.env.NODE_ENV !== 'production' && event && event.target) {
+        const targetType = event.target.type
+        const props: Object = this.props
+        const unknown =
+          ~['checkbox', 'radio', 'select-multiple'].indexOf(targetType) &&
+          !props.type
+
+        const type = targetType === 'select-multiple' ? 'select' : targetType
+        const { value }: any =
+          targetType === 'select-multiple' ? this.state.state || {} : props
+
+        if (unknown) {
+          console.error(
+            `Warning: You must pass \`type="${type}"\` prop to your Field(${
+              props.name
+            }) component.\n` +
+              `Without it we don't know how to unpack your \`value\` prop - ${
+                Array.isArray(value) ? `[${value}]` : `"${value}"`
+              }.`
+          )
+        }
+      }
+
       const value: any =
         event && event.target
           ? getValue(
@@ -151,8 +175,7 @@ export default class Field extends React.Component<Props, State> {
     } else if ((rest: Object).type === 'radio') {
       input.checked = value === _value
       input.value = _value
-    }
-    if (component === 'select' && (rest: Object).multiple) {
+    } else if (component === 'select' && (rest: Object).multiple) {
       input.value = input.value || []
     }
 
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -706,4 +706,68 @@ describe('Field', () => {
     expect(barValidate).toHaveBeenCalledTimes(1)
     expect(bazValidate).toHaveBeenCalledTimes(1)
   })
+
+  it('should warn when used without type prop and rendering radio, checkbox or multiple select indirectly', () => {
+    class Container extends React.Component {
+      render() {
+        return (
+          <Form initialValues={{ select_test: [] }} onSubmit={() => {}}>
+            {() => (
+              <div>
+                <Field name="checkbox_test" value="checkbox_value">
+                  {({ input }) => <input type="checkbox" {...input} />}
+                </Field>
+                <Field name="radio_test" value="radio_value">
+                  {({ input }) => <input type="radio" {...input} />}
+                </Field>
+                <Field name="select_test">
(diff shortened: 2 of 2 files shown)


erikras:
The `warning` function already does the `env` check. You can remove that.

erikras:
Can you add a test? Warnings are tested like [this](https://github.com/final-form/react-final-form/blob/master/src/Field.test.js#L79-L96).

erikras:
This is a great PR. I wonder if this warning shouldn't be in `getValue.js` since that's where the issue it's addressing is?

Andarist:
@erikras I was just commenting about env check, here it goes:

Each `warning` call should be wrapped with explicit `NODE_ENV` check for better minification, uglifyJS and other minifiers are not always able to dead code eliminate things, you can inspect your minified umd bundle and you will see that log messages are still there

let's consider even really simple example of
```js
// input.js
const warning =
  'production' === 'production'
    ? (condition, message) => {}
    : (condition, message) => {
        if (!condition) {
          console.error(`Warning: ${message}`)
        }
      }

warning(false, 'Log')
```

```js
// uglifyjs input.js -c --toplevel
((condition,message)=>{})(!1,"Log");
```

I've also explained this lately in those 2 comments ([here](https://github.com/dollarshaveclub/postmate/pull/55#discussion_r171670788) && [here](https://github.com/dollarshaveclub/postmate/pull/55#issuecomment-370606874)) if you want to read slightly more.

My recommendation is to wrap each call site to remove all this dead code. I can do it as part of this PR or I can remove this "explicit" env check from this PR and prepare a follow up PR wrapping each call site.

> Can you add a test? Warnings are tested like this.

Gonna look into it in a moment and ping u when I'm done with it.

> I wonder if this warning shouldn't be in getValue.js since that's where the issue it's addressing is?

I've briefly thought about it, but `getValue` doesn't have access to `props` and it was easier to check this and construct more helpful message here. Also this problem affects `render` first but we cannot warn there because we don't know what user will render and to what kind of DOM element it will hook the handlers.

> This is a great PR.

Thanks! I thought you are gonna like it :wink:

erikras:
> Each `warning` call should be wrapped with explicit `NODE_ENV` check for better minification

Awesome, I love being proved wrong! Makes total sense. I definitely wouldn't mind a PR converting all the other usages of `warning`...hint...hint... 🤔 😃

erikras:
...actually, now that I think about it, this revelation kind of renders the `warning` function totally useless. 😆

SimenB:
This is great! We were unable to get radios working properly with a custom component before sending in `type` to `Field`. Awesome with a great warning 🙂

src/Field.js
```diff
@@ -100,6 +100,29 @@ export default class Field extends React.Component<Props, State> {
...
+
+      if (process.env.NODE_ENV !== 'production' && event && event.target) {
+        const targetType = event.target.type
+        const props: Object = this.props
+        const unknown =
+          ~['checkbox', 'radio', 'select-multiple'].indexOf(targetType) &&
+          !props.type
+
+        const type = targetType === 'select-multiple' ? 'select' : targetType
+        const { value }: any =
```
Andarist: aint sure if this is entirely correct, if not - let me know what should be changed about it

src/Field.js
```diff
@@ -151,8 +174,7 @@ export default class Field extends React.Component<Props, State> {
     } else if ((rest: Object).type === 'radio') {
       input.checked = value === _value
       input.value = _value
-    }
-    if (component === 'select' && (rest: Object).multiple) {
+    } else if (component === 'select' && (rest: Object).multiple) {
```
Andarist: I think those are exclusive things, right?

src/Field.test.js
```diff
@@ -706,4 +706,68 @@ describe('Field', () => {
...
+    TestUtils.Simulate.change(radio, {
+      target: { type: 'radio', value: 'radio_value' }
+    })
+    TestUtils.Simulate.change(select, {
+      target: { type: 'select-multiple', value: ['select_value'] }
+    })
+
+    expect(spy).toHaveBeenCalled()
+    expect(spy).toHaveBeenCalledTimes(3)
+    expect(spy.mock.calls[0][0]).toBe(
```
Andarist: I find snapshots really nice for those kind of tests ("warning" text/errors thrown snapshots), but you do not use any snapshots so I wasn't sure if you'd like this approach.

Andarist:
@erikras provided a test, please take a look :)

> Awesome, I love being proved wrong! Makes total sense. I definitely wouldn't mind a PR converting all the other usages of warning...hint...hint... 🤔 😃

Sure thing, gonna send additional PR in a minute.

> ...actually, now that I think about it, this revelation kind of renders the warning function totally useless. 

It still encapsulates branching logic, not much - but might be nice to keep it. We'll discuss this under upcoming PR.

src/Field.js
erikras: Hopefully no one wants a `<select type="radio">`. 🤣

src/Field.test.js
erikras: I only just recently understood snapshot testing. Not actively against it, but not excited to migrate any tests that could be more concise with snapshots... 🤷‍♂️

erikras:

PR #206, opened 2018-03-28
Wrap each `warning` call site with NODE_ENV check for better dead cod…

Andarist:
What we've discussed under https://github.com/final-form/react-final-form/pull/204


erikras:
Having a whole module just to do an `if()` statement doesn't make sense to me. We're not even saving the `"Warning: "` string from the bundle size with the dead code removal. Plus, whenever I try to use `warning()`, I can never remember if `true` means to print the error or if `false` means to print the error.

Let's just move the `if` statements into each call site. Sound good?

Andarist:
Sure thing - done.

erikras:

PR #209, opened 2018-03-30
Provide field actions to <Field /> child

### LINKED ISSUE #179: Reset (or change the value of) a field at field-level

feature request (it exists at the form-level, but is not practical for some use cases)

Currently, I don't think there is any solution to reset a field (e.g. an input with a clear button). I tried:

```
clearField = (e) => {
  e.preventDefault();
  this.inputRef.value = '';
};
```

but the previous value came back on blur.

Maybe a `reset` or `change` (a la redux-form) prop passed to Field?

### Sandbox Link

https://codesandbox.io/s/6vm34649nn

### What's your environment?

RFF version: 3.1.2
FF version: 4.2.1
OS: MacOS High Sierra v10.13.3
Node version: 8.9.3

### Other information

Thanks for the work put into RFF!

Zephir77167:
Hey,

I would like to add a feature to the Field component, so it could perform actions on itself. Indeed, sometimes the form is not around and the field can't reset itself, or change value, blur, etc.

Please let me know if I could do things differently, or if I should add this behavior to other components as well for consistency. I'm not very familiar with react-final-form (yet!), so I would appreciate any guidance on this.

Thanks a lot!

Edit: fixes #179

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -143,7 +143,7 @@ export default class Field extends React.Component<Props, State> {
     }
   }
 
-  const fieldActions = {
+  fieldActions = {
     blur: () => {
       if (this.context.reactFinalForm) {
         this.context.reactFinalForm.blur(this.props.name);
--- a/src/Field.js
+++ b/src/Field.js
@@ -143,6 +143,29 @@ export default class Field extends React.Component<Props, State> {
     }
   }
 
+  const fieldActions = {
+    blur: () => {
+      if (this.context.reactFinalForm) {
+        this.context.reactFinalForm.blur(this.props.name);
+      }
+    },
+    change: (value: any) => {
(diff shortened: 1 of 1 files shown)


erikras:
`props.input.onChange(newValue)` will change the value of the field. And `onBlur` and `onFocus` do the same thing.

It's true that `reset()` is not available.

Zephir77167:
Ah I see, didn't think I could do it that way! Do you want me to update the PR to only include `reset()`? If so, should I put it in `props.input`? Else, you can just close the PR 👍

erikras:
No. `reset()` is just `props.input.onChange(props.meta.initial)`. 😄

Zephir77167:
Absolutely. Thanks a lot for pointers!

PR #219, opened 2018-04-10
Change FieldRenderProps['meta']['initial'] to any

aaronjensen:
Looks like this was mistyped.


erikras:
Heh. That same mistake was in the flow types. Thanks!

aaronjensen:
Sure thing. Do you know if there's an easy way to augment our types locally or do we have to wait for a new release?

PR #222, opened 2018-04-12
Generate externals dynamically from pkg.json

Andarist:
no actual change now, just preventive measure in case u add some deps in the future


erikras:
Interesting. By "preventive measure", what is it preventing? Me adding a dep but forgetting to put it into the rollup.config.js?

erikras:

PR #223, opened 2018-04-12
Use babel loose mode

Andarist:
pre 3003
post 2890


erikras:

PR #224, opened 2018-04-12
Added FormApi to Form and FormSpy props. Deprecated old.

### LINKED ISSUE #221: Expose FormApi

Feature request

Currently FormApi is accessible only in onSubmit, debug function (which is not executed in prod environment) and via decorators (am I'm missing something?).

My use case: in my app we have a form spread throughout several tabs and I need to highlight edited/invalid tabs. In order to do that I need to know field states, so FormApi.getFieldState().

Maybe we could just allow debug function to run on production environment and let the debug function decide whether it should be run in production?

erikras:
Fixes #221.


erikras:

PR #225, opened 2018-04-13
Fix regression: meta.data

PerfectPixel:
Adds `meta.data` back as it is listed as non-optional in the readme. 

Although it is flagged as optional in `final-form` it is always initialized with an object literal.

Updating to the newest `react-final-form` version also breaks the setFieldData-example: https://codesandbox.io/s/lrq5j4r8vl

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -169,6 +169,7 @@ export default class Field extends React.Component<Props, State> {
     const meta = {
       // this is to appease the Flow gods
       active: otherState.active,
+      data: otherState.data,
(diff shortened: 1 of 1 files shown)


erikras:
👍👍👍

PerfectPixel:
Wow that was super-quick 🥇 
thanks for the great library!

erikras:
Published in [`v3.3.1`](https://github.com/final-form/react-final-form/releases/tag/v3.3.1).

PR #229, opened 2018-04-19
Add meta.data to FieldRenderProps

PerfectPixel:


PerfectPixel:
I only deal in TypeScript and I missed this the first time around sorry 🙈

PerfectPixel:
But why is it optional, if the doc says its required?

erikras:
Great point. Everything in `FieldRenderProps` is optional, because maybe you're not subscribed to that part of the state. I'll fix the docs.

PR #236, opened 2018-04-23
Performance improvements

erikras:
Greatly improved first-render performance by pausing validation and notifications while mounting new fields.

Also upgraded deps.


erikras:
Published in [`v3.4.0`](https://github.com/final-form/react-final-form/releases/tag/v3.4.0).

PR #244, opened 2018-04-30
Add forceUpdate to FormRenderProps

perrin4869:
There are cases when form state will change, but there are no subscriptions available to be notified.
Such is the case with field data.
In such cases, having a `forceUpdate` function can be useful, although of course the use should be discouraged.
A quick and dirty, but maybe necessary workaround in some cases, until a better solution comes around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -700,6 +700,10 @@ _**[DEPRECATED]** Use `form.reset()` instead_
 A function that resets the form values to their last initialized values.
 [See the 🏁 Final Form docs on `reset`](https://github.com/final-form/final-form#reset-initialvalues-object--void).
 
+#### `forceUpdate: () => void`
+
+Forces rerender of the Form component
+
 ### `FormSpyProps`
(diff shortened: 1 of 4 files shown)


erikras:
This seems like API creep to me. It also seems like functionality that could be provided trivially by another component wrapping the form. No?

```jsx
    <UpdateForcer>
      {forceUpdate => (
        <Form onSubmit={onSubmit}>
          {({ handleSubmit }) => (
              Fields here
              <button onClick={forceUpdate}>Force Update</button>
            </form>
          )}
        </Form>
      )}
    </UpdateForcer>
```

perrin4869:
Hm... Well, at least at this point in `final-form` I have found a few cases where a `forceUpdate` would be useful (due to lack of notifications), so it would be useful to have that `forceUpdate` API built in, but you are right, the only justification I can bring for it is avoiding indentations.

PR #247, opened 2018-05-04
chore: do not install semantic-release on CI

DanielRuf:
semantic-release is not used on Travis CI builds so it is not neeeded to install it.


erikras:
It was aspirationally included. 😆

PR #248, opened 2018-05-04
chore: add Node.js 10

DanielRuf:
This adds Node.js 10 to the build matrix to make sure that the library also works on Node.js 10 (current stable).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ cache:
 notifications:
   email: false
 node_js:
+  - '10'
   - '8'
 script:
   - npm start validate


.travis.yml
```diff
@@ -1,11 +1,10 @@
 sudo: false
 language: node_js
-cache:
-  directories:
-    - node_modules
+cache: yarn
```
erikras: Can you explain this change?

DanielRuf: Well, yarn is not used on CI but npm, see https://travis-ci.org/final-form/react-final-form/builds/384784946#L450

erikras:
Thanks!

PR #250, opened 2018-05-08
Update Synchronous Field-Level Validation Example

levindixon:
Updated the existing example to use a more standard function composer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ form with no validation.
 Introduces a whole-record validation function and demonstrates how to display
 errors next to fields using child render functions.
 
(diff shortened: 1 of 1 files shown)


erikras:
Why is this better?

For one, it seems to apply them in reverse order, so the "must be a number" error is never shown.

levindixon:
Right to left/left to right composition flow can be controlled by using compose/flow (lodash/fp) or compose/pipe (rambdajs). By using a standard functional composition it allows you to follow all the practices of functional programming. It seemed like the example was trending towards using a functional programming approach, this update is just to get it all the way there. 

Could potentially distract from the main point the example is trying to make, if that's the case then feel free to close the PR 😄

PR #257, opened 2018-05-21
add allowReinitializeValues prop to fix nested values in initialValues causing reinitialization

### LINKED ISSUE #256: Nested values that change identity cause re-initialization

When you have nested values being passed into initialValues then recreating those initialValues causes a reset of the form back to its original value.

That `userStuff.firstName` would still be `"bob"`.

I am experiencing this because I have a large, complicated object that I'm editing in a form that has multiple small editors, each editing a small section of the big object. The current editor that you're looking at is controlled via the URL (hence the recreation with the `editingArea` prop in the example. My `initialValues` is generated from the object stored in an Apollo store, by doing some wrangling in order to get the object into the right shape for the form - hence the `getInitialValues` function in the example.

### Sandbox Link

https://codesandbox.io/s/6zmjz3nz8n

1. Change the first name from `"erikras"` to `"bob"` and see `userStuff.firstName` change to `"bob"`
2. Click the "Other Stuff" button
3. See the `userStuff.firstName` change back to `"erikras"`

### What's your environment?

🏁 React Final Form version: 3.4.2
🏁 Final Form version: 4.6.1 
OS: OS X 10.11.6
Browser: Chrome 66

bobbyrenwick:
This is merely a suggestion for how to fix #256, inspired by [formik's enableReinitialize](https://github.com/jaredpalmer/formik#enablereinitialize-boolean) and `[redux-form's enableReinitialize](https://redux-form.com/7.3.0/docs/api/reduxform.md/#-code-enablereinitialize-boolean-code-optional-). I'll admit that I just remembered the concept and didn't look at the names of those functions until just now, otherwise I probably would have chosen something a little more in-keeping!

I can change the name or the approach if you'd like. Just let me know what direction you'd like to take.

### DIFF
--- a/README.md
+++ b/README.md
@@ -612,6 +612,13 @@ to apply to the form.
 
 [See the 🏁 Final Form docs on `initialValues`](https://github.com/final-form/final-form#initialvalues-object).
 
+#### `allowReinitializeValues: boolean`
+
+A boolean that indicates whether or not a change in the shallow values of
+`initialValues` should cause the form to reinitialize. Setting this to `true` is
+useful when you are re-rendering the `Form` component, and you are using nested
+values.
+
 #### `mutators?: { [string]: Mutator }`
 
 [See the 🏁 Final Form docs on `mutators`](https://github.com/final-form/final-form#mutators--string-function-).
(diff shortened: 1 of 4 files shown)


erikras:
Closing this in favor of [workaround](https://github.com/final-form/react-final-form/issues/256#issuecomment-392564488).

PR #266, opened 2018-06-03
Add AsyncTypeahead and Redux example

perrin4869:
May be useful to someone trying to use the `AsyncTypeahead` along with `redux` and `react-final-form`. I am quite happy with this solution. Also, as far as I am concerned, this closes https://github.com/final-form/final-form/issues/122


erikras:
Very cool! 👍

PR #274, opened 2018-06-07
Added formatOnBlur prop

### LINKED ISSUE #187: Allow formatOnBlur

Feature Request

Formatting is immediate when user is doing an input. When the user is trying to edit a number with formatting, e.g., 1.5 percent is presented/formatted as 1.5 to user (value is stored as 0.015), the problem is when the user tries to backspace the number 1 (when the cursor is at the position before the dot in 1.5) and replace it with 3, the input becomes .5  (parsed and stored as 0.005), but then the formatting kicks in immediately, the .5 user input is replaced with 0.5, and the cursor goes to the end of the number. Then the user has to press left key two times, and then press 3. This gives a bad user experience since the user can't edit the text in the most natural way possible.

Another problem is when the user wanted to type `2.5`,  as soon as `2.` is typed, the value is formatted as `2`, this won't give user a chance to type `2.5` in the most natural way. The only way for `2.5` can be typed is by typing `25` first and then pressing the left key and then typing the `.` character.

Give an option that formatting happens only when the input loses focus.

Demo for the expected behavior. Backspace the number `1` in the `Interest Rate per Annum` of `1.5`, and then press `3`. http://jsfiddle.net/0zw630d4/4/

### Sandbox Link

Demo: https://codesandbox.io/s/nk1828jz70

### What's your environment?

Sandbox

erikras:
Fixes #187.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 ✅ Opt-in subscriptions - only update on the state you need!
 
-✅ 💥 [**2.9k gzipped**](https://bundlephobia.com/result?p=react-final-form) 💥
+✅ 💥 [**3.1k gzipped**](https://bundlephobia.com/result?p=react-final-form) 💥
 
 ---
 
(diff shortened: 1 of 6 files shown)


erikras:
Published in [`v3.6.0`](https://github.com/final-form/react-final-form/releases/tag/v3.6.0).

PR #272, opened 2018-06-06
Changed to not require RFF update when new FF config props are added

erikras:
Also adds `UNSAFE_` prefixes to fix #267.

Includes `react-lifecycles-compat` as described [on the React blog](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#open-source-project-maintainers).

PR #277, opened 2018-06-14
Actually use react-lifecycles-compat polyfill

erikras:
Fixes oversight from #272.

Installed in accordance with [instructions on React Blog](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#open-source-project-maintainers).

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react'
+import { polyfill } from 'react-lifecycles-compat'
 import PropTypes from 'prop-types'
 import { fieldSubscriptionItems } from 'final-form'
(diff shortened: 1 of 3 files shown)


jpdesigndev:
👍 Thanks! Are you cutting a new release for this?

erikras:
Published in [`v3.6.1`](https://github.com/final-form/react-final-form/releases/tag/v3.6.1).

erikras:
It turns out that the `react-lifecycles-compat` polyfill doesn't do what I thought it did. It only allows older versions of react to use libraries that use the _new_ lifecycle methods, like `getDerivedStateFromProps`, which RFF doesn't need or use. So I'm dumping it.

PR #287, opened 2018-06-25
Don't call setConfig for initialValues

dallonf:
This fixes a bug where re-rendering a form with a new `initialValues` object would reset all field values, even if the new `initialValues` were shallowly equal to the old ones. See the new test for an example of this.

This did cause a concerning change to another test, which was expecting `validate` to be called twice, but now it's only caused once. I think the latter behavior is actually correct and the test was written to work around this bug, but would appreciate some extra eyes on that test change to confirm.


erikras:
Published in [`v3.6.4`](https://github.com/final-form/react-final-form/releases/tag/v3.6.4).

PR #289, opened 2018-06-26
Fix support for nested forms by calling stopPropagation

### LINKED ISSUE #276: Support for Nested Forms?
Is it supported to have nested forms? For instance, a form where a field itself is a form.

I intend to make certain fields editable via a "popup" portal, where the popup itself is managed by a form. When you "submit" the nested form, the field updates.

However, it seems that the new FormState api is leaking into nested forms.

ineentho:
If you have two nested forms, both forms onSubmit were called because stopPropagation wasn't called. With this fix in place, it looks like nested forms are working properly.

Fixes #276.

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -101,9 +101,15 @@ class ReactFinalForm extends React.Component<Props, State> {
   }
 
   handleSubmit = (event: ?SyntheticEvent<HTMLFormElement>) => {
-    if (event && typeof event.preventDefault === 'function') {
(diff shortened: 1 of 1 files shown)


erikras:
Looks good. Can't hurt, I wouldn't think.

ineentho:
That was a fast response and merge! Thanks for your work!

erikras:
Published in [`v3.6.3`](https://github.com/final-form/react-final-form/releases/tag/v3.6.3).

PR #304, opened 2018-07-13
Changed result of handleSubmit() form render prop to Promise

martijnboland:
Small fix in TypeScript typings.


erikras:
Published in [`v3.6.5`](https://github.com/final-form/react-final-form/releases/tag/v3.6.5).

PR #309, opened 2018-07-29
Build separate bundle for React Native

Andarist:
If you like the change I'll have to adjust the tests - didnt want to do that now, because aint sure if you dig this 😉 

I'm also unsure if my changes are correct, would love you to look on them and validate it.

### DIFF
--- a/src/getValue.js
+++ b/src/getValue.js
@@ -16,18 +16,14 @@ const getValue = (
   event: SyntheticInputEvent<*>,
   currentValue: any,
   valueProp: any,
-  isReactNative: boolean
 ) => {
   if (
-    !isReactNative &&
     event.nativeEvent &&
     (event.nativeEvent: Object).text !== undefined
   ) {
     return (event.nativeEvent: Object).text
   }
-  if (isReactNative && event.nativeEvent) {
-    return (event.nativeEvent: any).text
-  }
+
   const detypedEvent: any = event
   const { target: { type, value, checked } } = detypedEvent
   switch (type) {
--- a/src/getValue.native.js
+++ b/src/getValue.native.js
@@ -1,0 +1,13 @@
+// @flow
+
+const getValue = (
+  event: SyntheticInputEvent<*>,
+  currentValue: any,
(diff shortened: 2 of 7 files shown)


src/getValue.js
```diff
@@ -16,18 +16,14 @@ const getValue = (
   event: SyntheticInputEvent<*>,
   currentValue: any,
   valueProp: any,
-  isReactNative: boolean
 ) => {
   if (
-    !isReactNative &&
     event.nativeEvent &&
```
Andarist: when `event.nativeEvent.text` is present in non-RN environment (I'm assuming browser target)?

src/getValue.native.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+const getValue = (
+  event: SyntheticInputEvent<*>,
+  currentValue: any,
+  valueProp: any,
+) => {
+  if (event.nativeEvent) {
```
Andarist: I've extracted this bit from the previous version of `getValue`, but it seems wrong to me, I have 0 experience with React Native though, so cant validate it :/

maybe this should be just
```js
const getValue = (event, currentValue, valueProp) => event.nativeEvent.text
```
?

erikras: I also have zero RN experience. Maybe we should find someone who does?

Andarist: cc @MichelDiz @rafaelcorreiapoli @geowarin @erickjth @bogdansoare @ibratoev @cosmith @maciejmyslinski @ghoshabhi

I think you guys were using (at least at some point in time) this library in RN. Could you take a look at this PR?

maciejmyslinski: I've never used react-final-form with react native. I asked my friends who code in react native for help, though.

Andarist: Cool! Sorry for pinging you then, I've looked through RN-related topics and pinged everyone who was involved.

crobinson42:
I have react-final-form working fine in react-native 0.60 on both Android and iOS.

PR #313, opened 2018-08-07
Migrate to new React Context API

doniyor2109:
I think it is good to migrate to new recommended Context API. There is also some library to integrate like `react-final-form-arrays`. I will also create PRs for them

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -1,6 +1,5 @@
 // @flow
 import * as React from 'react'
-import PropTypes from 'prop-types'
 import {
   configOptions,
   createForm,
@@ -14,11 +13,13 @@ import type {
   FormState,
   Unsubscribe
 } from 'final-form'
-import type { FormProps as Props, ReactContext } from './types'
+import type { FormProps as Props } from './types'
 import shallowEqual from './shallowEqual'
 import renderComponent from './renderComponent'
 import isSyntheticEvent from './isSyntheticEvent'
 import type { FormRenderProps } from './types.js.flow'
+import { ReactFinalFormContext } from './reactFinalFormContext'
+
 export const version = '3.6.0'
 
 const versions = {
@@ -39,17 +40,12 @@ type State = {
 }
 
 class ReactFinalForm extends React.Component<Props, State> {
-  context: ReactContext
   props: Props
   state: State
   form: FormApi
   mounted: boolean
   resumeValidation: ?boolean
   unsubscriptions: Unsubscribe[]
-
-  static childContextTypes = {
-    reactFinalForm: PropTypes.object
-  }
 
   constructor(props: Props) {
     super(props)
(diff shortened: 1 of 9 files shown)


erikras:
Is the HOC absolutely necessary?

doniyor2109:
@erikras I think it makes consumer reusable. What do you think?

erikras:
On closer examination, I think I have to agree. 👍

I'm on vacation this week, but I'll merge next week. In the meantime, why don't you modify this PR to export the HOC from this lib (changing both `index.js`, `index.js.flow`, and `index.d.ts`, and create a PR that imports it into `react-final-form-arrays`.

doniyor2109:
Sure I will add some works to this PR

src/reactFinalFormContext.js
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+
+export const ReactFinalFormContext = React.createContext({
+  form: null
+})
+
+export const withReactFinalForm = Component => {
+  return class extends React.Component {
+    render() {
+      return React.createElement(ReactFinalFormContext.Consumer, {
```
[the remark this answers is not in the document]
erikras: Great question. I've done this before because using JSX introduces, for little benefit, an extra build step / babel plugin, etc. JSX compiles to `React.createElement()` anyway.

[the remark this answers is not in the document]
Andarist: Sure, but it's used by the project either way & IMHO would make this slightly more readable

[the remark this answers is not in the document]
doniyor2109: @erikras Maybe it is time to move to JSX?

src/ReactFinalForm.js
```diff
@@ -281,13 +271,17 @@ class ReactFinalForm extends React.Component<Props, State> {
...
-    return renderComponent(
-      {
-        ...props,
-        ...renderProps,
-        __versions: versions
-      },
-      'ReactFinalForm'
+    return React.createElement(
+      ReactFinalFormContext.Provider,
+      { value: { reactFinalForm: this.form } },
```
Andarist: this is not flat, therefore breaking any shallow comparing underneath, or is it intentional to rerender each consumer when this renders?

doniyor2109: Where you want to apply shallow comparing?

doniyor2109: There is way to overcome such issue. I should write provide with reference value not with object literal

Andarist: well, its not exactly about the shallow comparing, but the principal is the same - https://reactjs.org/docs/context.html#caveats

doniyor2109: Yeah, I got your idea. Then I will change PR a little for this cavet

erikras:
This looks great! Thank you!

What does this do to the `react` peer dep in `package.json`? This is React v16-only, right?

erikras:
Obviously it's _eventually_ worth a breaking change. The question is: is that time now? 🤔

Andarist:
styled-components will soon make a similar change & many other libraries are also starting to rely on new context API. I would say it's about time - if one is still on <16.3 they can use a compatible react-final-form version instead of the newest one.

You could technically use https://github.com/jamiebuilds/create-react-context to polyfill context API, but that would be suboptimal for >=16.3 users as they already have "native" context.

erikras:
Published in [`v4.0.0`](https://github.com/final-form/react-final-form/releases/tag/v4.0.0).

PR #314, opened 2018-08-08
Pass form api via FieldRenderProps

doniyor2109:
Sometimes when developing forms, it is handful to get form api from inside filed component. I thought it would be great idea to access those api from field render props.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -226,7 +226,11 @@ class Field extends React.Component<Props, State> {
       // ignore meta, combine input with any other props
       return React.createElement(component, { ...input, children, ...rest })
     }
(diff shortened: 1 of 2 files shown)


erikras:
Can you update `index.d.ts` as well?

erikras:
How useful is this, really, since you're guaranteed to be inside the render function of the `<Form>`, which _does_ have access to this object? Can you describe a situation where this is useful?

doniyor2109:
@erikras I think this feature is can be solved with other solution. I released that Passing form API is only workaround. If user wants to make get form API, it is good to allow them all over the place not only limiting them on  `FieldRenderProps`. I think, the correct solution would be to use Context API and give `withForm` hos for user.

PR #350, opened 2018-10-10
Upgraded deps, including babel plugins

erikras:
Lowered bundle size by 100 bytes! 🎉


erikras:
Published in [`v3.6.7`](https://github.com/final-form/react-final-form/releases/tag/v3.6.7).

PR #351, opened 2018-10-10
Use @babel/runtime helpers instead of bundled ones

Andarist:
Focus of this PR is to use @babel/runtime helpers instead of including their copy in distributed sources.

**Motivation**
Shipping less code - by using @babel/runtime helpers we are able to deduplicate their code across all dependencies using them.


erikras:
This destroys my "ZERO DEPENDENCIES!!!" claim, though, does it not?

Andarist:
It depends on how do you look at it :P you already depend on this code, you just bundle it into your library instead of using external package 😉

erikras:
Published in [`v3.7.0`](https://github.com/final-form/react-final-form/releases/tag/v3.7.0).

Issue #181, opened 2018-03-04
Firefox, external submit doesn't prevent defaults and reload the page


leonardoanalista:
### Are you submitting a **bug report** or a **feature request**?
Bug report

### What is the current behavior?

Form is submitted and page is reloaded on Firefox. Tested on version 58.0.2, Mac.

### What is the expected behavior?

Form not be submitted and stays where is is without reloading the page.

### Sandbox Link

Example straight from the docs.
https://codesandbox.io/s/1y7noyrlmq

### What's your environment?

Bug is present in final form version 4.2.1, React-final-form 1.2.1, Firefox 58.0.2, Mac os.

### Other information

Make sure you click on button `External Submit via <code>document.getElementById()</code>
` to reproduce the bug.

DenisNeustroev:
the same, works fine in Chrome, fails in Firefox

petergombos:
I'm having the same problem, any workaround for this?

erikras:
Well, in the example, the "External Submit via closure" button works just fine. Is that not an acceptable workaround?

That the `document.getElementById()` method doesn't work seems more like a bug in React itself, namely that the `<form onSubmit>` is not being called.

leonardoanalista:
My workaround was pass a submit button as a prop to the Form component. Then Form component renders the submit button inside the form. Like this (removed private and irrelevant details):

```
export default class EditPlanDetails extends React.PureComponent {
  onClick = e => {
    const onSubmit = values => {
      this.props.saveData();
    };
        <DetailsForm
          onSubmit={onSubmit}
          initialValues={this.props.initialValues}
          submitButton={
            isReadOnly ? null : (
              <Clickable type="submit">OK</Clickable>
            )
          }
          cancelButton={ ... }
        />

  }

  render() {
    return (
      <Clickable
        onClick={this.onClick} >
           Show form Details
      </Clickable>
    );
  }
}
```

Then the form:
```
  render() {
    return (
      <Form
        onSubmit={this.props.onSubmit}
        initialValues={this.props.initialValues}
        validate={validate}
        render={({ handleSubmit, pristine, invalid, values }) => (
          <form id={this.props.id} onSubmit={handleSubmit}>
            <div>              
                <FinalField
                  name="tags"
                  label="Tags"
                  render={renderText}
                />              
            </div>
            <div className="text-right">
              {this.props.submitButton}
            </div>
          </form>
        )}
      />
    );
  }
}

```

[...]

PR #356, opened 2018-10-18
Fix FAQ external submit via document.getElementById() code example

chip4:
Code example in the FAQ now matches the code in the CodeSandbox example.
Using the code in the FAQ document before this change introduced an issue in Firefox addressed in: https://github.com/final-form/react-final-form/issues/181


erikras:
Thanks! 🙏

PR #366, opened 2018-10-30
Use loose mode for class properties transform

Andarist:
The main concern of this PR is to use `loose` mode for class properties transform. I took an opportunity to make a small babel config cleanup - instead of copying it to rollup.config.js and risking both it & babelrc being out of sync I've merged them both and put everything babel-related into `babelrc.js` file


erikras:
Published in [`v3.7.0`](https://github.com/final-form/react-final-form/releases/tag/v3.7.0).

Issue #374, opened 2018-11-15
Cannot create Field element because property reactFinalForm is missing


morloy:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

Flow gives an error when creating a form field:
`Cannot create Field element because property reactFinalForm is missing in props [1] but exists in object type [2].`

### What is the expected behavior?

No error.

### What's your environment?

```
    "final-form": "4.11.0",
    "react-final-form": "4.0.0",
```

erikras:
Published fix in [`v4.0.1`](https://github.com/final-form/react-final-form/releases/tag/v4.0.1).

morloy:
@erikras just noticed that the same happens with `FormSpy`.

erikras:
Duh. Sorry about that. Published fix in [`v4.0.2`](https://github.com/final-form/react-final-form/releases/tag/v4.0.2).

PR #377, opened 2018-11-16
Fixed flow FormSpy problem

erikras:
Fixes https://github.com/final-form/react-final-form/issues/374#issuecomment-439328156.


erikras:
Published in [`v4.0.2`](https://github.com/final-form/react-final-form/releases/tag/v4.0.2).

PR #386, opened 2018-11-26
ci: test Node.js 8, 10 and 11

DanielRuf:
Node.js 9 has reached its EOL. It is not a LTS branch like every even release branch. We should only test LTS branches. See https://github.com/nodejs/Release/blob/master/README.md


erikras:

PR #402, opened 2018-12-21
Externalize babel helpers using @babel/runtime

petermikitsh:
### tl,dr:

This reduces the ES module size from `27545` to `26852` bytes (~2.5%) by externalizing babel runtime helper functions.

Many people are using these helper functions elsewhere. By externalizing them, folks can take advantage of only including a single copy (as opposed to the additional copy that was inlined by `react-final-form`).

It appears this was the original intent of the repository since `@babel/runtime` is [already a production dependency](https://github.com/final-form/react-final-form/blob/master/package.json#L118).

### Details

[react-final-form.es.js](https://unpkg.com/react-final-form@4.0.2/dist/react-final-form.es.js) goes from something like this:

```js
import { createElement, Component, createContext } from 'react';
import { fieldSubscriptionItems, configOptions, createForm, formSubscriptionItems, version } from 'final-form';

function _extends() {
  _extends = Object.assign || function (target) {
    for (var i = 1; i < arguments.length; i++) {
      var source = arguments[i];

      for (var key in source) {
        if (Object.prototype.hasOwnProperty.call(source, key)) {
          target[key] = source[key];
        }
      }
    }

    return target;
  };

  return _extends.apply(this, arguments);
}

// Other babel helper functions (not shown).
```

to this:

```js
import _extends from '@babel/runtime/helpers/extends';
import _objectWithoutPropertiesLoose from '@babel/runtime/helpers/objectWithoutPropertiesLoose';
import _inheritsLoose from '@babel/runtime/helpers/inheritsLoose';
import { createElement, Component, createContext } from 'react';
import { fieldSubscriptionItems, configOptions, createForm, formSubscriptionItems, version } from 'final-form';

//      
function diffSubscription (a, b, keys) {
  if (a) {
    if (b) {
      // $FlowFixMe
      return keys.some(function (key) {
        return a[key] !== b[key];
      });
    } else {
      return true;
    }
  } else {
    return !!b;
  }
}
```

### Related

https://github.com/final-form/react-final-form-arrays/pull/60


Andarist:
Yeah, that was the intention implemented in https://github.com/final-form/react-final-form/pull/351 . However seems that I've screwed it up in https://github.com/final-form/react-final-form/pull/366 .

.babelrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   ],
   plugins: [
     '@babel/plugin-transform-flow-strip-types',
+    '@babel/plugin-transform-runtime',
```
Andarist: could you remove it from `.babelrc.js` and add it like this: https://github.com/final-form/react-final-form/pull/351/files#diff-ff6e5f22a9c7e66987b19c0199636480R86 ?

petermikitsh: @Andarist Fixed ✅

Andarist:
@erikras friendly 🏓

petermikitsh:
Fixed merge conflict, all checks pass

erikras:

PR #413, opened 2019-01-16
Update FieldProps.validate type definition

hayfield:
final-form#368 updated the README with the updated prototype for FieldProps.validate.
This change makes the associated update to the Typescript types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ that selects all of the items of
 wish to update for. If you don't pass a `subscription` prop, it defaults to
 _all_ of [`FieldState`](https://github.com/final-form/final-form#fieldstate).
 
-#### `validate?: (value: ?any, allValues: Object, meta: FieldState) => ?any`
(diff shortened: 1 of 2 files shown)


erikras:

StefanBilliet:
Hi

Could it be that the new typing info didn't make it in the 4.1.0 npm package? I just installed it and in index.d.ts I still see:

export interface FieldProps<T extends HTMLElement>
  extends RenderableProps<FieldRenderProps<T>> {
  allowNull?: boolean;
  format?: ((value: any, name: string) => any) | null;
  formatOnBlur?: boolean;
  parse?: ((value: any, name: string) => any) | null;
  name: string;
  isEqual?: (a: any, b: any) => boolean;
  subscription?: FieldSubscription;
  **validate?: (value: any, allValues: object) => any;**
  value?: any;
  [otherProp: string]: any;
}

PR #418, opened 2019-01-30
Setup TS tests

Andarist:
I've added infrastructure to test TS typings by possible use cases which should prevent regressions in the future (once more tests get added ofc)


erikras:
This pull request **introduces 2 alerts** when merging 90a471e27d56e5c9b191a03e317bb329f4b90749 into 65a239b27284d02ab7e0284b9dd7b9087514f97c - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-83c2e127d3dd79127cc188f3c316bb86af4f8a49)

**new alerts:**

* 2 for Unused variable, import, function or class

---

Andarist:
@erikras LGTM report is actually an expected result (those rules should be disabled for .ts files as their purpose is to test interfaces and not to actually be executed), but LGTM setting is not in this repo (I think) and thus I cannot fix it :<

erikras:
This pull request **introduces 2 alerts** when merging 029914a82461b99dc8e2e58e9e64f951e64ccf92 into 768a476cc2323d895f5b96447c147ee56b5bf793 - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-46eecb3fc27495f5ca09f0d0864a09758fd22203)

**new alerts:**

* 2 for Unused variable, import, function or class

---

Andarist:
@erikras new version of `dtslint` got published, this PR should be ready to merge now. Would be good to tweak LGTM setup to ignore rules such as `no-unused-vars` in `typescript` directory

erikras:

typescript/index.d.ts
```diff
@@ -0,0 +1,104 @@
...
+  FormState,
+  FormSubscription,
+  FieldSubscription
+} from 'final-form';
+
+export interface ReactContext {
+  reactFinalForm: FormApi;
+}
+
+export interface FieldRenderProps<T extends HTMLElement> {
```
pokonski: Hey @Andarist, can you tell why FieldRenderProps now has the an argument now?

Andarist: Truth to be told - I don't know. I don't recall making this change, but I've looked through git history and my local reflog and it seems I was indeed the one making it.

That being said the change makes sense to me `FieldRenderProps["input"]` is a set of closely related things, the generic should constraint the whole thing and not the specific handlers separately, because you could call them with different input type.

Could you paste in the example code which broke by this change? Could you describe why do you think this change should be reverted? We could now start writing tests to be sure such changes won't affect existing codebases.

However it's somewhat a known problem (and I dont think we can really improve on that as community) for libraries shipping type definitions - personally I dont think that typings change should require a major bump, it would be too restrictive. So the best approach is to lock your dependencies and update frequently.

pokonski:
I just started using the library and noticed some TypeScript examples of how people write their components for fields and neither had the argument prior to the latest version of the react-final-form. 

Is the argument really needed if it's not used for anything inside the interface?

pokonski:
For example https://github.com/insxne/acid-forms/blob/c011b55a682ce459e9f027af70fe97e5bda9bdbb/src/LoginForm/TextField.tsx#L5

Andarist:
> Is the argument really needed if it's not used for anything inside the interface?

But it's used very much so - in event handlers declarations, i.e. [here](https://github.com/final-form/react-final-form/blob/a3ca5f169da5cb5abcf53d60285d8d43b08c427f/typescript/index.d.ts#L24)

Check out what this generic parameter gives u - https://github.com/final-form/react-final-form/pull/439/files#diff-c9d158e9961d9803798bff495584efabR1 . I could potentially add a default but it would be `= HTMLElement`, dunno if that would be useful.

pokonski:
@Andarist ah indeed! It still feels a bit weird to have to specify it as an argument to those props.

For example Formik has it set to any:
https://github.com/jaredpalmer/formik/blob/02b8911f72554495afa18ddb87eaf1f4296d97ff/src/types.tsx#L140

as does ReduxForm:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/792f0a463bdc976ab1a22f405f183e2409df389d/types/redux-form/lib/Field.d.ts#L80

Andarist:
> @Andarist ah indeed! It still feels a bit weird to have to specify it as an argument to those props because it basically can be either an input or a select, right?

It could also be a textarea and probably (but not sure) a custom element (as in webcomponent). 

In my opinion TS's goal is to type check our programs and it can only do that if we give it appropriate & strict hints. Using `any` anywhere is the opposite of safety and IMHO we should strive to avoid it at all costs - personally I try to use it only sometimes in very local places, casting it back to something conrete as soon as possible.

We could provide both versions under the same name using conditional types but it seems weird to me to allow non-strict versions because it probably will get used more frequently than strict one due to lack of exposure.

That being said we have to take developer ergonomics into account, I'm far from TS expert and I don't exactly know what are conventions used currently by the community, so would love having more people discussing this.

PR #426, opened 2019-02-18
Add Form component FormData generic typing

michalwarda:
In TypeScript you can use generic syntax inside of JSX code to allow typing components.
`react-final-form` exposes a way to type `Config` `FormData` through `createForm` but there is no way to access it in `react-final-form`.

This pull request exposes the possibility to pass `FormData` typing to `Form` component.

I've also added typings tests for that.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -69,7 +69,9 @@ export interface RenderableProps<T> {
   render?: (props: T) => React.ReactNode
 }
 
-export interface FormProps extends Config, RenderableProps<FormRenderProps> {
+export interface FormProps<FormData = object>
+  extends Config<FormData>,
+    RenderableProps<FormRenderProps> {
   subscription?: FormSubscription
   decorators?: Decorator[]
   initialValuesEqual?: (a?: object, b?: object) => boolean
@@ -94,7 +96,9 @@ export interface FormSpyProps extends RenderableProps<FormSpyRenderProps> {
 }
 
 export var Field: React.ComponentType<FieldProps>
-export var Form: React.ComponentType<FormProps>
+export class Form<FormData = object> extends React.Component<
+  FormProps<FormData>
+> {}
 export var FormSpy: React.ComponentType<FormSpyProps>
 export var version: string
 
--- a/src/ReactFinalForm.d.test.tsx
+++ b/src/ReactFinalForm.d.test.tsx
@@ -116,3 +116,29 @@ function mutated() {
     </Form>
   )
 }
+
+// with typed form data
+
+const typedOnSubmit = (values: { firstName: string }) => {
+  console.log(values)
+}
+
+function withTypedFormData() {
+  return (
+    <Form<{ firstName: string }> onSubmit={typedOnSubmit}>
+      {({ handleSubmit }) => (
+        <form onSubmit={handleSubmit}>
+          <div>
+            <label>First Name</label>
+            <Field
+              name="firstName"
+              component="input"
+              type="text"
+              placeholder="First Name"
+            />
+          </div>
+        </form>
+      )}
+    </Form>
+  )
+}


erikras:
This pull request **introduces 1 alert** when merging 657125c349f036ded453afc6ad210830c00af8b4 into 768a476cc2323d895f5b96447c147ee56b5bf793 - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-a0025b928e21938c546da1605f4c1ca9ce506322)

**new alerts:**

* 1 for Unused variable, import, function or class

---

wtgtybhertgeghgtwtg:
`values` and `initialValues` will still be loosely typed.  Should those get updated, as well?

michalwarda:
@wtgtybhertgeghgtwtg I think that would require changing the type definitions inside `final-form` [here](https://github.com/final-form/final-form/blob/master/src/index.d.ts#L242) and [here](https://github.com/final-form/final-form/blob/master/src/index.d.ts#L53).
I'll be glad to help with that but I'm not sure how to approach this multi-repo change.

wtgtybhertgeghgtwtg:
I'd guess make them generic with a default in a minor release for `final-form`, then bump the version in `peerDependencies` for the next `react-final-form` major.

erikras:
@michalwarda I'd accept the PR for the FF changes, and then we can merge this RFF change, just like @wtgtybhertgeghgtwtg said.

@wtgtybhertgeghgtwtg I'm curious about the sentence that your username is abbreviating.

wtgtybhertgeghgtwtg:
It's not an abbreviation, I just mashed the keyboard.

michalwarda:
@erikras I'll do it in few hours :)

whengely:
Are the changes going into FF the blocker currently for this or is it something else?

src/index.d.ts
```diff
@@ -69,7 +69,9 @@ export interface RenderableProps<T> {
   render?: (props: T) => React.ReactNode
 }
 
-export interface FormProps extends Config, RenderableProps<FormRenderProps> {
+export interface FormProps<FormData = object>
+  extends Config<FormData>,
+    RenderableProps<FormRenderProps> {
   subscription?: FormSubscription
   decorators?: Decorator[]
   initialValuesEqual?: (a?: object, b?: object) => boolean
```
stramel: Shouldn't these also be `FormData`?

src/index.d.ts
```diff
@@ -94,7 +96,9 @@ export interface FormSpyProps extends RenderableProps<FormSpyRenderProps> {
 }
 
 export var Field: React.ComponentType<FieldProps>
-export var Form: React.ComponentType<FormProps>
+export class Form<FormData = object> extends React.Component<
```
danielnixon: `Form` is actually implemented as a function, not as a class, so this type could be misleading couldn't it?

It would suggest I can create instances a la `const form = new Form()`?

src/index.d.ts
```diff
@@ -94,7 +96,9 @@ export interface FormSpyProps extends RenderableProps<FormSpyRenderProps> {
 }
 
 export var Field: React.ComponentType<FieldProps>
```
danielnixon: It would be nice for `Field` to get the same treatment.

erikras:
Can you resubmit this PR (or resolve the conflicts)?

src/index.d.ts
Jephuff: I've been doing this if it helps. 
```
<FormData = object>(
  props: FormProps<FormData>,
) => React.ReactElement<FormProps<FormData>>
```

erikras: [It's `React.FC` now](https://github.com/final-form/react-final-form/blob/master/typescript/index.d.ts#L102).

erikras:
Thank you for this contribution! It provided guidance and impetus to getting this feature into the library.

I think #516 did what this PR did, so I think this one is obsolete now.

PR #427, opened 2019-02-19
re-write to solely use commit-phase lifecycle methods

Sharakai:
## "Discussion" (and potential implementation) PR as to how best to make `react-final-form`:
- `React` 16.(9?)+ compliant
- Work properly with `React.StrictMode` (already out)
- Work properly with `React.ConcurrentMode` (~Q2 2019)
(dates taken from https://reactjs.org/blog/2018/11/27/react-16-roadmap.html)

### About `React.StrictMode` and `React.ConcurrentMode`
`React.StrictMode` runs only in `NODE_ENV === 'development'`, and calls all render-phase lifecycle methods **twice** - This includes the `constructor`. This is because these lifecycles should be pure, as `React`'s "Concurrent Mode" makes no guarantees that render-phase methods are only called once. In fact, they can be called as many times as needed, and may never be committed at all.

### About the `ReactFinalForm` double-render (or how I learned to love the `return null`)

I'm using a double-render trick with `componentDidMount` to cut out an initial (pointless) final-form-not-ready render, so the flow looks like this:
1. Render-phase lifecycle methods called (twice)
2. `render` returns `null`, as it isn't mounted yet
3. `componentDidMount` called, which sets up the form, does all the initialisation, and calls `setState({ mounted: true });`
4. `render` is called again from the `setState`, and now actually renders the form (with the `final-form` form).

The initial `render` returning null is **not** committed to the DOM, so the user will only see the final result
> You may call setState() immediately in componentDidMount(). It will trigger an extra rendering, but it will happen before the browser updates the screen. This guarantees that even though the render() will be called twice in this case, the user won’t see the intermediate state.
>
> &mdash; _https://reactjs.org/docs/react-component.html#componentdidmount_

### Next steps - Counter-PR

Hopefully within the next week or two, I'll raise a counter-PR with another way to do it, by ensuring the render-phase lifecycle methods (primarily, the `constructor`) can handle being called twice, but only do anything if `this.form` hasn't already been constructed for that class instance. I will still have to remove deprecated lifecycle methods, though (e.g. `componentWillMount`)

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -3,7 +3,11 @@ import * as React from 'react'
 import { fieldSubscriptionItems } from 'final-form'
 import diffSubscription from './diffSubscription'
 import type { FieldSubscription, FieldState } from 'final-form'
-import type { FieldPropsWithForm as Props, FieldRenderProps } from './types'
+import type {
+  FieldPropsWithForm as Props,
+  FieldRenderInputProp,
+  FieldRenderMetaProp
+} from './types'
 import renderComponent from './renderComponent'
 import isReactNative from './isReactNative'
 import getValue from './getValue'
@@ -15,12 +19,13 @@ const all: FieldSubscription = fieldSubscriptionItems.reduce((result, key) => {
 }, {})
 
 type State = {
-  state: ?FieldState
+  state: ?FieldState,
+  mounted: boolean
 }
 
 class Field extends React.Component<Props, State> {
   props: Props
-  state: State
+  state: State = { state: undefined, mounted: false }
   unsubscribe: () => void
 
   static defaultProps = {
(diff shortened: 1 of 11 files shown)


Sharakai:
Closing this as it's been open for 18 days with no comments / reviews, and is now (unsurprisingly) conflicting on every file.

Without changing over to using commit-phase lifecycle methods, this library will be unusable once react concurrent mode comes into play, and is already difficult to use in `NODE_ENV === 'development'` mode with `React.StrictMode` due to the double-invocation of all render-phase methods (and `constructor`).

Happy to re-do this after discussion - Feel free to email me or comment on this PR.

erikras:
Sometimes I don't get the time to read every PR for a few weeks. OSS is not my job. Closing the PR before any discussion or review is a good way to ensure that I don't see it. I just happened to notice this today by chance.

Your discussion looks thoughtful. If you know of a competing way to do it, I'd like to see that PR as well, so we can consider together how to best approach this.

erikras:
This should be fixed in `v5`. Reopen if not.

PR #429, opened 2019-02-20
Add Typescript typings for the third argument of the `validate` function

ecrofeg:
This is where the third argument of the `validate` function has been added: https://github.com/final-form/final-form/pull/125
But no Typescript typings were added.
Fixed.

Also, I've moved `meta` typings into separate `type`, to avoid typings duplication.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -12,6 +12,25 @@ export interface ReactContext {
   reactFinalForm: FormApi
 }
 
+export type FieldPlaneState = Partial<{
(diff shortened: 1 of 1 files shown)


ecrofeg:
Ok, looks like `FieldState` interface already exists in `final-form` typings, so I used that instead.
Furthermore, I've excluded all functions from that interface.

ecrofeg:
@erikras can we have it please? No serious changes.

erikras:

PR #433, opened 2019-03-04
Lock handleSubmit while submitting

Andarist:
Actually this probably could be included in final-form itself. If you'd like to include this change, let me know in which package and I'll update my PR (or move it) appropriately. I might also write tests once you approve the logic

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -44,6 +44,7 @@ class ReactFinalForm extends React.Component<Props, State> {
   state: State
   form: FormApi
   mounted: boolean
+  submitLock: boolean
   resumeValidation: ?boolean
   unsubscriptions: Unsubscribe[]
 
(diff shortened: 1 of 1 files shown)


erikras:
So this is just `submit() { if(submitting) { return } ..... }` logic?

erikras:
...and yes, FF is the place for this.

Andarist:
> So this is just submit() { if(submitting) { return } ..... } logic?

Yeah, total brain fart, lol. I've kinda copy pasted what I've put into my app to handle this (as I dont get easy access to the FF state from outside of the <ReactFinalForm/>). Totally missed the obvious fact that I can actually use the FF state here 😅

erikras:
Published fix in [`final-form@4.13.0`](https://github.com/final-form/final-form/releases/tag/v4.13.0).

PR #438, opened 2019-03-06
Fix input.onChange TS type signature

strattadb:
Hi!

The Flow type signature of `input.onChange` is [`(event: SyntheticInputEvent<*> | any) => void`](https://github.com/final-form/react-final-form/blob/master/src/Field.js#L117) while the TypeScript signature is [`onChange: (event: React.ChangeEvent<T>) => void`](https://github.com/final-form/react-final-form/blob/master/typescript/index.d.ts#L25).

This PR adds the `| any` in the TS definition.

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -22,7 +22,7 @@ export interface FieldRenderProps<T extends HTMLElement> {
   input: {
     name: string;
     onBlur: (event?: React.FocusEvent<T>) => void;
-    onChange: (event: React.ChangeEvent<T>) => void;
+    onChange: (event: React.ChangeEvent<T> | any) => void;
     onFocus: (event?: React.FocusEvent<T>) => void;
     value: any;
     checked?: boolean;


circlingthesun:
Thanks, I was just about to submit a PR for this.

typescript/index.d.ts
```diff
@@ -22,7 +22,7 @@ export interface FieldRenderProps<T extends HTMLElement> {
   input: {
     name: string;
     onBlur: (event?: React.FocusEvent<T>) => void;
-    onChange: (event: React.ChangeEvent<T>) => void;
+    onChange: (event: React.ChangeEvent<T> | any) => void;
```
Andarist: You are not adding any as possible argument type, you are rather *changing* it to any.

If you make union with any you basically make the whole thing `any`, you can verify this by pasting in `type A = string | any` into TS playground and inspecting A's type.

Why would you like to make this change? Could you paste in your usage example to demonstrate how it fails with current typings?

circlingthesun: Have a look at the tests, there are many examples where onChange is called with a string or number type. For example: https://github.com/final-form/react-final-form/blob/master/src/ReactFinalForm.test.js#L339

Andarist: IMHO its an example of badly written test - it ain't testing the real thing. It hijacks argument passed to the `renderInput` and calls it from the outside with completely different argument type (string) from the one it would be called with in a real environment. It's just confusing for the reader.

That being said you are right - it seems that `onChange` can be indeed called with `any` which is IMHO a mistake - it acts as a setter in disguise then (if not called with an event). Would be better to have other explicit API for setting value like this (there is one but it's not provided to the value)

So to sum this up - we should actually change the argument type for onChange to just `any` because as mentioned `| any` makes it `any` anyway.

We could maybe provide later stricter typings (disallowing `any` here) based on some conditional types, but that would be a subject for another PR.

Also - as there are now tests for typings, it would be good to add one that would cover this or at least notify people in the future that ability to pass any value to onChange is required in one form or another.

circlingthesun: When using the onChange prop on dom elements like input, select, or textarea the React.ChangeEvent<T> argument signature makes 100% sense. However, many custom input components don't follow the same convention (see https://github.com/JedWatson/react-select for an example). You can't really enumerate all the possible types, so I'm not sure disallowing any makes sense.

Andarist: Well - in the final-form it would actually be possible to enumerate all possible types, it would require some runtime breaking changes though due to the current design. Disallowing `any` is always a good thing (when it makes sense ofc) and in most cases it is possible (especially with the help of generics)

Andarist: Additional thought after the night - types don't necessarily have to cover all "valid" runtime patterns, they are here to prevent footguns and this sometimes mean that they should disallow certain patterns.

Take React as example - there is `forwardRef` which accepts a function accepting props & ref, where ref is defined as
```ts
type Ref<T> = { bivarianceHack(instance: T | null): void }["bivarianceHack"] | RefObject<T> | null;
```
Let's take closer look at one of this union - `RefObject`. It's defined as:
```ts
interface RefObject<T> {
  readonly current: T | null;
}
```

1. Is it true that `current` is readonly at runtime? **No.**
2. Can I find a use case where I need to write to the `current` property? **Definitely yes.**
3. Should the typing allow writing to the current `property` here? **Probably no.**

circlingthesun: You make a good argument here. I don't however think that it applies to restricting onChange's `event` argument to React.ChangeEvent<T>. If generics can somehow be used to narrow it down to React.ChangeEvent<T> or <typeof value> I'm all for it. Otherwise people using react-select and other components will have to force it to `any`, or painstakingly wrap values in synthetic events.

Andarist: Could you prepare a simple usage example with react-select so I could examine it to see how types flow through it?

onlyann: I use the Ant Design Checkbox component with react-final-form that wraps the [rc-chekbox](`https://github.com/react-component/checkbox`).

The restricted signature of the `onChange` prop prevents me from upgrading.

I am with @circlingthesun that some components don't necessarily expose on `onChange` events matching a `React.ChangeEvent<T>` signature. 

It is great that the `Field` component works with `any` event and the TypeScript definition should reflect it.

onlyann: An alternative solution could be to define the `Field` in the following manner

```typescript
export interface FieldRenderProps<T = HTMLInputElement, TChangeEvent = React.ChangeEvent<T>> {
  input: {
    name: string;
    onBlur: (event?: React.FocusEvent<T>) => void;
    onChange: (event: TChangeEvent) => void;
    onFocus: (event?: React.FocusEvent<T>) => void;
    value: any;
    checked?: boolean;
  };
  meta: FieldPlaneState;
}

export interface FieldProps<T = HTMLInputElement, TChangeEvent = React.ChangeEvent<T>>
  extends RenderableProps<FieldRenderProps<T, TChangeEvent>> {
  allowNull?: boolean;
  format?: ((value: any, name: string) => any) | null;
  formatOnBlur?: boolean;
  parse?: ((value: any, name: string) => any) | null;
  name: string;
  isEqual?: (a: any, b: any) => boolean;
  subscription?: FieldSubscription;
  validate?: (value: any, allValues: object, meta?: FieldPlaneState) => any;
  value?: any;
  [otherProp: string]: any;
}

declare class Field<T = HTMLInputElement, TChangeEvent = React.ChangeEvent<T>> extends React.Component<FieldProps<T, TChangeEvent>, FieldState> {
}

```

Usage:
```typescript
// field render props
function fieldRenderProps() {
  return (
    <Form onSubmit={onSubmit}>
      {({ handleSubmit, reset, submitting, pristine, values }) => (
          <Field name="firstName">
            name="firstName"
            placeholder="First Name"
          >{({ input }) => {
            return <input {...input} />;
          }}
          </Field>
          <Field<HTMLSelectElement>
            name="title"
            placeholder="Title"
          >{({ input }) => {
            return <select {...input} />;
          }}
          </Field>
          <button
            type="button"
            onClick={reset}
            disabled={submitting || pristine}
          >
            Reset
          </button>
          <button type="submit" disabled={submitting || pristine}>
            Submit
          </button>
          <pre>{JSON.stringify(values)}</pre>
        </form>
      )}
    </Form>
  );
}


// custom component
function customComponent() {
  interface Props {
    onChange: (value: string) => void;
    onFocus: () => void;
    onBlur: () => void;
  }

  class MyComponent extends React.Component<Props> {}

  return (
    <Form onSubmit={onSubmit}>
      {({ handleSubmit, reset, submitting, pristine, values }) => (
          <Field<string, string>
            name="firstName"
            placeholder="First Name"
          >{({ input }) => {
            return <MyComponent {...input} />;
          }}
          </Field>
          <button
            type="button"
            onClick={reset}
            disabled={submitting || pristine}
          >
            Reset
          </button>
          <button type="submit" disabled={submitting || pristine}>
            Submit
          </button>
          <pre>{JSON.stringify(values)}</pre>
        </form>
      )}
    </Form>
  );
}

```

onlyann:
#441

goldins:
Hi, what's the status on this? I would like this change as well. As far as I can tell, it's simply reverting the definition back to what it was in v4.0.2: https://github.com/final-form/react-final-form/blob/v4.0.2/src/index.d.ts#L19

typescript/index.d.ts
PerfectPixel: Another use case would be datepickers or their fallbacks. Native `input[type=date]` are not supported in Safari or IE. One could implement a fallback proposed by mozilla at [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Examples). There, one input is replaced by three selects in the fallback. In that component you need to compute the actual value and pass it to `react-final-form`s onChange.

General discussions about `any` being a good idea aside. This PR is just about bringing the typescript types back in line with the actual library function, its flow types and the documentation. Should this not be merged to fix the regression that was introduced and then - if `any` should be disallowed - done in a regular fashion and released as a breaking change?

erikras:
Whether or not it was a good decision, the API has always supported `onChange` being given _either_ a React Event or a value. There _are_ some valid use cases where this comes in handy, e.g. the [Declarative Form Rules](https://github.com/final-form/react-final-form#declarative-form-rules) example.

onlyann:
Hey @erikras , any chance a new minor release can be pushed to NPM for the TypeScript fix?

PR #443, opened 2019-03-13
expose form name to field and withReactFinalForm.

vospascal:
dependend on pull request in https://github.com/final-form/final-form/pull/204 in final-form

I would like to add support for name property/attribute to final-form since form name is a default property/attribute, its handy to get some context when dynamically and re-using custom field components.
This is part one of a two part pull request the other one will be for react-final-form.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-final-form",
-  "version": "4.0.2",
+  "version": "4.1.0",
   "lockfileVersion": 1,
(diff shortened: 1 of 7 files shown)


vospascal:
i dont know wy travis is failing on some node versions could use some help there

vospascal:
See for usefull discussion https://github.com/final-form/final-form/pull/204
For code example see here https://codesandbox.io/s/w248n2z8j7?fontsize=14

PR #447, opened 2019-03-24
Update readme that warnings do not prevent submit

Noitidart:
I think this is a correction. If it is not, please reject.


erikras:
The original way can be read correctly where "that prevents submission" modifies "error", but your wording is definitely more clear.

PR #449, opened 2019-03-28
[Typescript] Fix input.onChange signature

HenriJ:
`input.onChange` accepts events but also directly values as per the documentation.
The typescript definition was not in sync with this behaviors


erikras:
Dupe of #438.

PR #460, opened 2019-04-26
Update README.md

gtournie:
Add redux-form-validators code sandboxes


erikras:
@gtournie Can you fix your "alphanumeric" validation in the sandbox?

```js
      with: /^[a-zA-Z0-9 ]$/i,
```
should be:
```js
      with: /^[a-z0-9 ]*$/i,
```
(no need for `A-Z` because it's already case insensitive)

erikras:
@gtournie ping.

gtournie:
@erikras Fixed. Sorry for the delay.

erikras:
[Not for `username`](https://codesandbox.io/s/y43o4vqj1).

PR #463, opened 2019-05-01
Fix StrictMode errors

### LINKED ISSUE #319: Warnings in strict mode
I just wrapped my react.js app in a <[`StrictMode`](https://reactjs.org/docs/strict-mode.html)> component, and react-final-form yielded the following warnings, two of which are complaining about react-final-form. 

> Warning: Unsafe lifecycle methods were found within a strict-mode tree:
>     in App
> 
> componentWillMount: Please update the following components to use componentDidMount instead: Autocomplete, **ReactFinalForm**
> 
> componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Autocomplete, Transition, t
> 
> componentWillUpdate: Please update the following components to use componentDidUpdate instead: **ReactFinalForm**
> 
> Learn more about this warning here:
> https://fb.me/react-strict-mode-warnings
> 

I assume that this will repro in any app that's using react-final-form wrapped in a `<StrictMode>` component, but if not let me know and I'll build a simplified repro case.

stramel:
Fixes #319 

Attempted to maintain the existing functionality with the non-deprecated lifecycle methods. Although, I think #427 might be a better long term approach, it doesn't seem to have activity from the author.

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -121,6 +121,7 @@ class ReactFinalForm extends React.Component<Props, State> {
         this.resumeValidation || !this.form.isValidationPaused()
       this.form.pauseValidation()
     }
+    return null
   }
 
   componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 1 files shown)


erikras:
This should be fixed in `v5`. Reopen if not.

PR #465, opened 2019-05-10
Use updated value for `formatOnBlur` format func

michaltk:
When there's a `format` function that's called `onBlur` it passes the
current value from state to that `format` function.
https://github.com/final-form/react-final-form/blob/41d24d6db197a5eec73e52623a267f98bb2337dc/src/Field.js#L113

This causes that when the field's `onChange` is called from within the `onBlur`
the `formatOnBlur` function is not getting the updated value.
Rather, it has the previous value passed to it, so when
formatting the value it ends up formatting and returning the previous value.

This pr changes to pass the current updated value from `props.reactFinalForm`'s state,
to the format func so it always gets the current value.
[See Demo](https://codesandbox.io/s/j2j8w0nlp5)

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -278,6 +278,43 @@ describe('Field', () => {
     expect(getByTestId('name').value).toBe('ERIKRAS')
   })
 
+  it('should `formatOnBlur` most updated value', () => {
+    const format = jest.fn(value => (value ? value.trim() : ''))
+    const handleBlur = event => {
+      fireEvent.change(getByTestId('name'), {
+        target: { value: event.target.value.toUpperCase() }
+      })
+    }
+    const { getByTestId } = render(
+      <Form onSubmit={onSubmitMock} subscription={{ values: true }}>
+        {() => (
+          <form>
+            <Field name="name" format={format} formatOnBlur>
+              {({ input }) => (
+                <input
+                  {...input}
+                  data-testid="name"
+                  onBlur={e => {
+                    handleBlur(e)
+                    input.onBlur(e)
+                  }}
+                />
+              )}
+            </Field>
+          </form>
+        )}
+      </Form>
+    )
+    const inputText = '   erikras'
+    fireEvent.focus(getByTestId('name'))
+    expect(getByTestId('name').value).toBe('')
+    fireEvent.change(getByTestId('name'), { target: { value: inputText } })
+    expect(getByTestId('name').value).toBe(inputText)
+    fireEvent.blur(getByTestId('name'))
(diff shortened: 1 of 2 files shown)


erikras:
Things have changed with v5, but if this is still a problem, we can address it.

I'm not understanding your concern. In your demo, you should just be doing the `toUpperCase().trim()` in the `format` function. I'm not understanding why you would want to call `onChange` first?

michaltk:
I've updated to v5 and this still seems to be an issue.
[See updated small demo of where this is needed.](https://codesandbox.io/s/w0qbn)
We have a stand alone email input component that shows the suggested domain while typing.
It then updates the email to the suggested domain `onBlur` by calling the `onChange` with the suggestion. (This is a standard form field that an work with any form).
We are wrapping this input with final-form and would like to trim the email input `onBlur` by passing `formatOnBlur` to the final-form Field.
The issue is that the `format` function still have the value as the email without the suggested domain
so it formats that and the suggestion never gets set.

erikras:
I've implemented what I understand your requirements to be:

1. No whitespace around email address (`trim()`)
1. Autocomplete should work
1. Product name should turn to all uppercase on blur

Did I miss any?

[Edit 🏁 React Final Form - Format On Blur Example](https://codesandbox.io/s/react-final-form-format-on-blur-example-r2029?fontsize=14)

michaltk:
With the example, it wouldn't allow spaces to be typed.
For a states auto complete, that would not work. We would need to allow typing a space and therefore would have to have the format only run onBlur.

[See updated state auto complete demo](https://codesandbox.io/s/091cb)

michaltk:
@erikras Thanks for checking into this!
I've updated this pr to work with v6..

erikras:
I don't doubt that this fixes your problem, but I'd like to have a test that demonstrates the problem and fails with this code commented out, but passes with it included. Either you can do it, or I can try to get around to it later this week...

erikras:
Your test is pretty good, but it's not, the parlance of the React Testing Library motto, "how a user would use the library". Primarily: `handleBlur` would not have access to "fire event" on the input DOM node.

Andarist:
Why, why, why you'd want to call imperatively onChange inside the onBlur? this sounds like a misusage to me, but maybe I'm just not aware about the use case - would be good to describe it 🤔

erikras:
I agree that this feels _wrong_ somehow. I'll have to give more thought to the motives behind this change. It also created an untested code branch. I'm not sure when the ternary on line 115 would ever flip the other way.

erikras:
Published in [`v6.3.0`](https://github.com/final-form/react-final-form/releases/tag/v6.3.0).

PR #467, opened 2019-05-13
Hooks!!!

erikras:
This is a _complete rewrite_ of 🏁 React Final Form (obviously much of the logic was saved), away from ES6 classes and lifecycle methods, to use only [React Hooks](https://reactjs.org/docs/hooks-intro.html). Believe it or not, this provides even better control of how, and in which stages of the render cycle, 🏁 React Final Form communicates with its internal 🏁 Final Form instance.

It also includes a migration of _all_ of the tests to use [🐐 React Testing Library](https://testing-library.com/docs/react-testing-library/intro), which is a superior testing environment.

Late one insomniatic night, it occurred to me that all of the hooks logic that was internal to `Field` and `FormSpy` could be exported for public consumption, so it was all abstracted out into `useField` and `useFormState` hooks respectively.


erikras:
This pull request **introduces 1 alert** and **fixes 1** when merging de77bb3b6764dc39733d25ec100b318ec00bcc6c into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-f494938c47751f2bc8a9465176d7bb077f6368af)

**new alerts:**

* 1 for Unused variable, import, function or class

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

README.md
```diff
@@ -479,6 +474,18 @@ A component that takes [`FormProps`](#formprops) and surrounds your entire form.
...
+
+The `useField` hook takes two parameters, the first is the name of the field, and the second is an optional object that looks just like [`FieldProps`](#fieldprops), except without the name. It returns an object just like [`FieldRenderProps`](#fieldrenderprops).
+
+`useField` is used internally inside `Field`.
+
+### `useFormState`
+
+The `useFormState` hook takes one optional parameter, which matches the exact shape of [`FormSpyProps`](#formspyprops). It returns a [`FormSpyRenderProps`](#formspyrenderprops).
+
+`useFormState` is used internally inside `FormSpy`.
```
Andarist: you are not renaming FormSpy after all? :p

src/Field.js
```diff
@@ -1,240 +1,63 @@
...
-      // ignore meta, combine input with any other props
-      return React.createElement(component, { ...input, children, ...rest })
-    }
-    const renderProps: FieldRenderProps = { input, meta } // assign to force Flow check
-    return renderComponent(
-      { ...renderProps, children, component, ...rest },
-      `Field(${name})`
-    )
-  }
+Field.defaultProps = {
```
Andarist: IMHO it's better to just use es6 default params, especially that components with defaultProps might even trigger slower paths inside of react occassionally, so... use the platform 😉 

additionally this is a static property and as I've mentioned this breaks a possibility to tree-shake Field 😢

src/FormSpy.js
```diff
@@ -1,194 +1,44 @@
...
-            reactFinalForm.reset(eventOrValues)
-          }
+  const state = useFormState({ onChange, subscription })
+  const renderProps: FormSpyRenderProps = {
+    form: {
+      ...reactFinalForm,
+      reset: eventOrValues => {
+        if (isSyntheticEvent(eventOrValues)) {
+          // it's a React SyntheticEvent, call reset with no arguments
+          reactFinalForm && reactFinalForm.reset()
```
Andarist: `reactFinalForm` cant be undefined here, I think Flow should understand that - but might be wrong

src/ReactFinalForm.js
```diff
@@ -34,63 +36,128 @@ export const all: FormSubscription = formSubscriptionItems.reduce(
...
-    }
-    this.unsubscriptions = []
-    if (this.form) {
-      // set initial state
+  // synchronously register and unregister to query form state for our subscription on first render
+  const [state, setState] = React.useState<FormState>(
+    (): FormState => {
       let initialState: FormState = {}
-      this.form.subscribe((state: FormState) => {
+      form.subscribe(state => {
```
Andarist: maybe this is an opportunity for a getState on the form? or do you want to discourage using it as part of the public API?

src/ReactFinalForm.js
```diff
@@ -34,63 +36,128 @@ export const all: FormSubscription = formSubscriptionItems.reduce(
...
         initialState = state
       }, subscription || all)()
```
Andarist: does this (`subscription || all`) matter here? you unsubscribe right away anyway

src/ReactFinalForm.js
```diff
@@ -34,63 +36,128 @@ export const all: FormSubscription = formSubscriptionItems.reduce(
...
-    if (this.mounted) {
-      this.setState({ state })
+  const flattenedSubscription = flattenSubscription(subscription || all)
+  React.useEffect(() => {
+    // We have rendered, so all fields are no registered, so we can unpause validation
+    form.isValidationPaused() && form.resumeValidation()
+    const unsubscriptions: Unsubscribe[] = [
+      form.subscribe(s => {
+        if (firstRender.current) {
+          if (validationStateChanged(state, s)) {
```
Andarist: is it the only important thing that we can get here which should trigger rerender? 🤔

src/ReactFinalForm.js
```diff
@@ -34,63 +36,128 @@ export const all: FormSubscription = formSubscriptionItems.reduce(
...
+    decorators,
+    () => {
+      console.error(
+        'Warning: Form decorators should not change from one render to the next as new values will be ignored'
+      )
+    },
+    // this env check short circuts the comparison in production
+    // so we don't run the shallowEqual on every render
+    // istanbul ignore next
+    process.env.NODE_ENV === 'production' ? () => true : shallowEqual
```
Andarist: u could also wrap this `useWhenValueChanges` in env check to ensure that this string etc goes away correctly in production, you can break the hooks rules if u know what you are doing 😉

src/useConstant.js
```diff
@@ -0,0 +1,21 @@
...
+ * the lifetime of the component
+ *
+ * Plagiarized from https://github.com/Andarist/use-constant
+ *
+ * @param {Function} init - A function to generate the value
+ */
+export default function useConstant<T>(init: () => T): T {
+  const ref = React.useRef<Maybe<T>>(undefined)
+  if (!ref.current) {
+    ref.current = init()
```
Andarist: just a note - I on purpose wrap this in an object and return its property so you technically can return falsy value from init and things still work as expected

src/useField.js
```diff
@@ -0,0 +1,205 @@
...
+  // istanbul ignore next
+  if (process.env.NODE_ENV !== 'production' && !reactFinalForm) {
+    console.error(
+      'Warning: useField must be used inside of a ReactFinalForm component'
+    )
+  }
+
+  // keep ref to most recent copy of validate function
+  const validateRef = React.useRef(validate)
+  validateRef.current = validate
```
Andarist: IMHO this is not really the proper way of saving latest value, it's slightly better to use such logic as here https://github.com/Andarist/use-latest/ 

the difference is subtle and probably not really relevant, but I believe the other way of doing this is more predictable - gonna come back with explanation 2morrow

src/useField.js
```diff
@@ -0,0 +1,205 @@
...
+  //   initialValue,
+  //   isEqual,
+  //   validateFields,
+  //   reactFinalForm
+  // })
+  React.useEffect(
+    () =>
+      register(state => {
+        if (firstRender.current) {
+          firstRender.current = false
```
Andarist: in theory the state might change between 1st render and this effect, you probably should conditionally setState if it has changed in between

src/useField.js
```diff
@@ -0,0 +1,205 @@
...
+          setState(state)
+        }
+      }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [
+      name,
+      defaultValue,
+      // If we want to allow inline fat-arrow field-level validation functions, we
+      // cannot reregister field every time validate function !==.
+      // validate,
```
Andarist: is this comment still relevant - https://github.com/final-form/react-final-form/pull/467/files#diff-04c6e90faac2675aa89e2176d2eec7d8R574 ? by keeping validate in ref (which you do), you can guarantee that validate can be inlined and it can also be conditional and still wont cause reregister

src/useWhenValueChanges.js
```diff
@@ -0,0 +1,21 @@
...
+    if (!isEqual(value, previous.current)) {
+      callback()
+    }
+    // We really only care if value changes, as callback doesn't change.
+    // This would not be safe to release as a custom hook for public
+    // consumption, but because we know the specific uses in this library,
+    // we know that we can ignore changes to isEqual and callback
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [value])
```
Andarist: u are doing custom compare inside, so IMHO this dep should be removed

src/useWhenValueChanges.js
```diff
@@ -0,0 +1,21 @@
...
+  value: any,
+  callback: () => void,
+  isEqual: (any, any) => boolean = (a, b) => a === b
+) {
+  const previous = React.useRef(value)
+  React.useEffect(() => {
+    if (!isEqual(value, previous.current)) {
+      callback()
+    }
+    // We really only care if value changes, as callback doesn't change.
```
Andarist: seems to me that you have forgotten about:
```js
previous.current = value
```

src/useField.js
Andarist: Consider what might in theory happen - react might render at arbitrary times (and thus updating this ref) and in theory `validate` might change each time, so when `validate` get called it's one of those functions, but we have no idea which one.

If we update it in useEffect (after commit phase) then we at least know which version of that function it is, the last one from before commit phase.

Moving it to useEffect also corresponds nicely with how classes work, what you want here is `this.props.validate`, but because of heavy closure usage in hooks it's not as simple as reaching `props.validate`, but conceptually this is what you want, right? So in class worls `this.props` gets updated during commit phase, kinda just before componentDidUpdate.

README.md
erikras: Sometimes the vociferous minority in a vote can sway the outcome. 😜

src/Field.js
erikras: Excellent point. I overlooked this. 👍

src/FormSpy.js
erikras: Good catch! It [_was_ undefined-able (?) once](https://twitter.com/erikras/status/1127989923334754315), but I engineered a way around it.

erikras: Well, it _can_ be if the developer is a doofus and tries to put a `<FormSpy>` outside of a `<Form>`, hence the warning above. Flow is in the right here.

src/ReactFinalForm.js
erikras: It _could_ be, but I think it should respect the subscription. It shouldn't give you all of the form state on first render and then to have it restricted to your subscription on subsequent renders.

Great thought, though.

src/ReactFinalForm.js
erikras: It's about what data gets put into the state. See other comment. https://github.com/final-form/react-final-form/pull/467#discussion_r283662032

src/FormSpy.js
Andarist: Oh, I thought you throw error above - wouldnt it be better to throw as it's an obvious mistake and you wouldnt have to guard against missing value all over the place?

src/ReactFinalForm.js
erikras: I'm 99% sure, yes. The only thing that happens between start of form render and end of form render is that fields get registered, and the only form state that changes when that happens is potentially validation via field-level validation functions.

Timeline looks like this:

1. `Form` first render
   1. create `form`
   1. register/unregister to get initial state
   1. throw `form` into context
   1. pause validation!
   1. call render prop
1. `Field`s first render
   1. register/unregister with form to get initial state
   1. call render prop
1. `Field`'s `useEffect` fires
   1. create lasting subscription to `form`
   1. register field-level validation functions
1. `Form`'s `useEffect` fires
   1. validation unpaused and all registered field-level validation functions run
   1. create lasting subscription to `form`
   1. the only part of form state that could've changed is its validity due to the field-level validation functions

src/ReactFinalForm.js
Andarist: Oh, thought subscription specifies which keys should trigger a callback, didnt know it also specifies which exact keys you are going to receive. Makes sense that way then.

src/ReactFinalForm.js
Andarist: In pure theory i.e. FormSpy can change an arbitrary thing between render & commit - so in theory something else might change

src/ReactFinalForm.js
erikras: GREAT point! 👍

src/useConstant.js
erikras: Yes. Yours is very general purpose in a way that I don't need mine to be. I'll add a comment.

src/useField.js
erikras: I left the `// validate` commented to show future me / others that including it was, indeed, considered, but a decision was intentionally made to not include it.

If you mean something else, say it again.

src/useWhenValueChanges.js
erikras: Great point. 👍

src/useWhenValueChanges.js
erikras: 🤦‍♂️ Hooray for code review!! Will fix.

README.md
```diff
@@ -564,6 +571,8 @@ _all_ of [`FieldState`](https://github.com/final-form/final-form#fieldstate).
 A function that takes the field value, all the values of the form and the `meta` data about the field and
 returns an error if the value is invalid, or `undefined` if the value is valid.
 
+⚠️ IMPORTANT ⚠️ – By default, in order to allow inline fat-arrow validation functions, the field will not rerender if you change your validation function to an alternate function that has a different behavior. If you need your field to rerender with a new validation function, you will need to update another prop on the `Field`, such as `key`.
```
Andarist: My comment [here](https://github.com/final-form/react-final-form/pull/467#discussion_r283669309) was more about this IMPORTANT note rather the other code line. This note here is kinda confusing to me - is `validate` updatable? As in - will the "latest" validate be used or just the first one? 

If it's updatable when knowing that it doesn't cause a rerender would matter? Also what wouldnt rerender in this case?

src/FormSpy.js
erikras: Good point. I decided to make `ReactFinalForm` throw errors; so I'm going to make them all throw if you're using it incorrectly, production or not.

src/ReactFinalForm.js
erikras: Shit. You're right. Hence my 99%...

This was an attempt to not render twice when field level validation is used (that passes), but maybe that's just gotta be the price of using field-level validation.

Thanks for challenging me on these things. 👍 Add another 🍺 to your earnings when we meet.

erikras: Well.... if I disable this check, the form has to render twice _always_, regardless of your validation choices...

...what if we make the rule: **if you want to modify form state on first render in a `FormSpy`, you're gonna need to put it in a `setTimeout` to trigger it later.** ?

src/useField.js
erikras: Gotcha! Great explanation! 👍

src/useWhenValueChanges.js
erikras: 👍

src/useField.js
erikras: Not sure I understand...

README.md
erikras: Maybe this will clear it up? If you don't include [this line here](https://github.com/final-form/react-final-form/blob/hooks/src/Field.test.js#L455), the field won't rerender again as invalid after swapping out the validation function.

It _is_ updatable, but if you need a rerender, you've gotta change a prop that will cause it. Make more sense? Is there a better way to word the IMPORTANT note?

erikras:
This pull request **introduces 1 alert** and **fixes 1** when merging 1abd7f3debc9dfc36fed9799d55255ea2457c66d into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-176ef0abdb40bf87e8a4eff1c4eb7bf237f55cf9)

**new alerts:**

* 1 for Unused variable, import, function or class

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

src/useField.js
Andarist: From what I understand `register` gives `state` synchronously upon registration (just like `subscribe`).

So this is a classic hooks-subscription problem. Technically the state can change between your first register in lazy initialization setState and subscription point in `useEffect`.

README.md
Andarist: Hm, ok - I get it. Not sure though if this has to be said - it sounds complex, but in reality you shouldnt use `validate` inside of your function-child, so you indeed doesnt have to rerender when it changes. OTOH it's obviously good to mention/document every edge case. 🤷‍♂ 

Maybe something like:
> You shouldn't rely on `validate` inside of your component, just let the Final Form handle that for you. If you really need to rerender `Field` along with `validate` change then force the rerender by a new value of some other prop.

src/useField.js
Andarist: Just a question for now - seems slightly weird that `validate` gets special cased like this. Is there no other prop that should get such treatment as well? (or none should?)

src/Field.js
```diff
@@ -1,240 +1,58 @@
...
-      this.state.state && this.state.state.focus()
-    }
+import useField, { defaultFormat, defaultParse } from './useField'
+
+const Field = ({
+  allowNull,
+  children,
+  component,
+  defaultValue,
+  format = defaultFormat,
```
Andarist: as those are passed directly to `useField` and people might use `useField` directly without ever using `Field` it seems that those defaults should be handled in the `useField` itself

src/useField.js
```diff
@@ -38,27 +38,26 @@ const useField = (
...
       'Warning: useField must be used inside of a ReactFinalForm component'
     )
   }
 
   // keep ref to most recent copy of validate function
   const validateRef = React.useRef(validate)
-  validateRef.current = validate
+  React.useEffect(() => {
+    validateRef.current = validate
+  }, [validate])
```
Andarist: the effect itself is a simple assignment, so it really shouldnt be costful, I would skip the dep (`[validate]`) altogether here, if `validate` stays the same then it will also stay the same on the ref, even after assignment

src/useFormState.js
```diff
@@ -0,0 +1,55 @@
...
+          setState(newState)
+          if (onChange) {
+            onChange(newState)
+          }
+        }
+      }, subscription || all),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    flattenedSubscription
+  )
+  return onChange ? undefined : state
```
Andarist: this conditional return is slightly weird, at least at first - could you explain briefly why this is handled like this?

src/Field.js
erikras: I put them there, but then `input.value` was getting formatted to `''` when `undefined`, so I took it back out. Not sure what's best... 🤔

src/useField.js
erikras: Yeah, I flip-flopped on that when writing it. I'll remove it.

src/useFormState.js
erikras: Because it was copied from `FormSpy`, which specifically doesn't render anything if you're using `onChange`, but it's stupid this way, so I'll make it return the state. Good catch!

src/useField.js
erikras: > From what I understand `register` gives `state` synchronously upon registration (just like `subscribe`).

Yes. That's correct.

How do I
> conditionally `setState` if it has changed in between

? I'm not 100% on what you mean by that.

erikras:
This pull request **introduces 1 alert** and **fixes 1** when merging 2f2ca8a323fb5ef4ee1b42ffd1582504bba01bd8 into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-37e2d0503952d4c330318a48f3c28b91abfcd61f)

**new alerts:**

* 1 for Unused variable, import, function or class

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

src/useField.js
Andarist: if its first render you must compare receives state against the one in react's state and if they differ you should setState (thus triggering "double render", but it is what it is)

erikras: I've been trying to avoid including `deepEqual` code in the library. Perhaps it's unavoidable?

Andarist: why cant this be based on shallowEqual? which part of the state is not comparable as easily?

src/ReactFinalForm.js
Andarist: Gonna give it a thought some time later hopefully

src/useField.js
erikras: Same here. Gonna go think about something else and see if any insights arrive. 😄

erikras:
This pull request **fixes 1 alert** when merging 9a7125bc0c342cd6c675a320724b3820ab49a07e into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-cf9e04a837481f8a1d9d4588ed1574ecb88a28d4)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

erikras:
This pull request **fixes 1 alert** when merging c1408c5f0b6d3516301be7cd2cf0fb22e8c4090d into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-85a3e2b6c7e0e875e949572d6c7617112e591b34)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

src/ReactFinalForm.js
Andarist: Cant we just do a shallow check here? this would cover all differences and from what I see in the source code state values which are not primitives, but objects, are having identities preserved by reusing `lastFormState.objectValue` if shallow equal pass.

erikras: Okay, so the keys that change are the objects `modified`, `touched`, and `visited`, which, with a simple form with one `name` field, go from `{}` to `{ name: false }` on first render.

We could either:

1. special case ignore those for triggering a rerender, or
1. leave it how it is, only checking validation-related keys, or
1. just render all the damn forms twice at first

Thoughts? 🤔

erikras: On reflection, it doesn't _have_ to be "all the damn forms", because you can select to not subscribe to those maps. But by default, it should use `shallowEqual` to compare the new and old state, and if you don't provide a subscription, your form is just gonna render twice at first.

erikras:
This pull request **fixes 1 alert** when merging 4870a834d3206aa59f5515309d21f213e795def7 into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-586a3e523269e7207b7d5c739a77fa473731c062)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

erikras:
This pull request **fixes 1 alert** when merging 0b786abc2542ec7d86cf8068bb2ba9c0baa7c457 into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-a188a4ec1f45e93a4a650a5f249df57c80a94eec)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

erikras:
This pull request **fixes 1 alert** when merging ef289e60f5372162fc04fc81002de8c4a61f66b1 into 41d24d6db197a5eec73e52623a267f98bb2337dc - [view on LGTM.com](https://lgtm.com/projects/g/final-form/react-final-form/rev/pr-decee0ec0bddd29289f2b99729703f26b8f7b6ec)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

src/ReactFinalForm.js
Andarist: Seems fine to me 👌

PR #477, opened 2019-05-15
Bug Fix: Add state.change to useField's onChange dependencies.

jkdowdle:
First off, I love the project and appreciate being able to use it.

In my project I have what may be an unconventional form where the fields that are rendered on the screen depend on what the user selects. The form as well as all the selectable categories are visible at the same time.  I found a problem when I was migrating to version 5.x in my unit tests where once a subset was 'selected' a second time a textarea onChange event no longer updated the fields value. 

I was trying to debug my own code and found no problems so I started playing with the code under the hood and I found that `state.change` was not part of the dependencies of the onChange useCallback. I added it and my test started passing as expected. 

Adding state.change as a dependency is all the pull request does, but if there is a reason that state.change should not be in the dependencies (maybe performance?) then I could work on creating a reproducible example.

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -140,7 +140,7 @@ const useField = (
         state.change(parse ? parse(value, name) : value)
       },
       // eslint-disable-next-line react-hooks/exhaustive-deps
-      [_value, name, parse, state.value, type]
+      [_value, name, parse, state.change, state.value, type]
     ),
     onFocus: React.useCallback((event: ?SyntheticFocusEvent<*>) => {
       state.focus()


erikras:
That is very odd. The value of `state.change` is _only_ ever being set to [this](https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L751), which only contains the `name` bound via closure. Are you changing the `name` prop to your `<Field>`? We've got some name-changing tests, but maybe they're not robust enough...

jkdowdle:
Let me take another look at my test and make sure the name is changing, it should be and I am pretty sure it is. 

I did think that was odd as well because I know name is in the dependency array. I will attempt to reproduce it and take a look at those tests. Hopefully I will have something soon.

jkdowdle:
@erikras 

Ok here is reproducible example. I had some odd behavior with Codesandbox, so here is a repo. I actually had to add more elements than I expected but it still should be easy to follow. 

https://github.com/jkdowdle/react-final-form-issue

```
git clone https://github.com/jkdowdle/react-final-form-issue.git 
cd react-final-form-issue
yarn 
yarn test
```

I first tried to use react-testing-library rerender utility to change to a different field name prop, but that did work as expected. I haven't looked at the unit tests yet, but I would imagine it was similar to that?

Also notice the difference between the modified form state.

erikras:
Published in [`v5.1.1`](https://github.com/final-form/react-final-form/releases/tag/v5.1.1).

jkdowdle:
Thanks a ton @erikras!

PR #478, opened 2019-05-16
Used final-form@4.13's new beforeSubmit callback to enable formatOnBlur on submit

### LINKED ISSUE #470: formatOnBlur: Also run on submit
I'm using `formatOnBlur` to trim strings. This works fine, but if someone submits by pressing the enter key without leaving the field first, then no blur event happens and the string is submitted as-is without the formatter code running.

Or is there a better way to trim all strings before submission? format+formatOnBlur is pretty convenient because we're using a custom Field wrapper component that always sets it for string fields.

erikras:
Fixes #470.


erikras:
Published in [`v5.1.0`](https://github.com/final-form/react-final-form/releases/tag/v5.1.0).

PR #486, opened 2019-05-20
Adjust TypeScript definitions

ksocha:
Adjust TypeScript definitions to match v5 of the lib.


erikras:
Thanks! I was certain I'd missed something on the TS side. 👍

erikras:
Published in [`v5.1.1`](https://github.com/final-form/react-final-form/releases/tag/v5.1.1).

PR #494, opened 2019-05-25
Slightly improve TypeScript types

danielnixon:


danielnixon:
A small change that would allow me to do what I need to do with `Config` in lieu of the real fix coming in https://github.com/final-form/react-final-form/pull/426.

PR #495, opened 2019-05-26
Fix `submitting` state never getting reset

### LINKED ISSUE #487: `Submitting` not getting reset after submit
When  subscribed to `submitting` state, `submitting` never gets updated back to `false` even after submitting.
This seems to work fine in v4 and only is an issue in v5. 

After submit, `submitting` state should be `false` again.

### Sandbox Link
[See Demo](https://codesandbox.io/s/react-final-form-simple-example-7slu8)

### What's your environment?
React Final Form - 5.1.0
Final Form - 4.13.0

### LINKED ISSUE #492: Form doesn't rerender when going back from dirty state to pristine state
**bug report**

Form doesn't rerender when going back from dirty state to pristine state

Form should rerender when going back from dirty state to pristine state

### Sandbox Link
https://codesandbox.io/s/r9e0k

1. Change value from 'bla' to some other value
2. You'll see 'Is pristine: false'
3. Change value back to 'bla'
4. You'll see that 'Is pristine' is still false
5. Click 'get state' button
6. In the form state in console you'll see that pristine is true

### Other information
Works fine in react-final-form@4.1.0

michaltk:
fixes #487

Previously, when it was checking in the subscribe func if state updated https://github.com/final-form/react-final-form/blob/4ca1ce7a7800c08b249219e14740ec2c1ed7e222/src/ReactFinalForm.js#L89  the current state `state` - was always the initial state, bec of js closures.

So, the `submitting` state got set to `true` bec that was different than the initial state but when attempting to set back to false.. it was checking the initial state of `false` with current updated state of `false` and it was equal and therefore did not update and reset state.

Changed to add `state` to the `useEffect` deps so it will rerun the effect, causing it to create a new subscribe func with the updated state.

fixes #492

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -103,7 +103,7 @@ const ReactFinalForm = ({
       unsubscriptions.forEach(unsubscribe => unsubscribe())
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [decorators, flattenedSubscription])
+  }, [decorators, flattenedSubscription, state])
 
   // warn about decorator changes
   // istanbul ignore next


erikras:
This is wrong. We can't unsubscribe and resubscribe every time state changes.

I'm reverting in another commit and fixing the problem. I've created a test to reproduce it.

erikras:
Published my fix for this in [`v6.0.0`](https://github.com/final-form/react-final-form/releases/tag/v6.0.0).

PR #496, opened 2019-05-27
Changed parse and format API to appease Flow gods

### LINKED ISSUE #400: Can not set parse or format to null with Flow

Bug report

Flow gives an error for form fields with parse or format set to `null`
`Cannot create Field element because:
 • null [1] is incompatible with function type [2] in property parse.`

No errors

### What's your environment?

"react-final-form": "4.0.2",
"final-form": "4.11.0",
"flow-bin": "0.84",

### Other information

Probably the fix is to slightly change type defs (it works in my project, but it has only one form so I am not sure about possible? side effects) 
```
  format?: ((value: any, name: string) => any) | null,
  parse?: ((value: any, name: string) => any) | null,
```

erikras:
Fixes #400.

Flow will not allow a union type of `Function | null`, which is how the `parse` and `format` API was designed. That was a terrible design. **This is breaking change**. To achieve what was previous achievable with passing `null` to `parse` or `format` can now be achieved by passing the identity function, `v => v`.


src/useField.js
```diff
@@ -180,8 +180,10 @@ const useField = (
     visited: otherState.visited
   }
   if (formatOnBlur) {
-    value = defaultFormat(value, name)
-  } else if (format) {
+    if (component === 'input') {
```
ThiefMaster: This breaks cases where `component` is a custom adapter: #514

ThiefMaster: @erikras any chance you could have a look at this? it's currently preventing us from updating from 4.1 right now :/

PR #498, opened 2019-05-27
Fix form rerender bug

erikras:
This is the proper fix for #487 and #492. Reverts #495.


erikras:
Published in [`v6.0.0`](https://github.com/final-form/react-final-form/releases/tag/v6.0.0).

PR #499, opened 2019-05-27
Update Types to make useField's options optional

rockingskier:
`useField` has a [default value](https://github.com/final-form/react-final-form/blob/master/src/useField.js#L39) but the Types do not.

That means `useField('name')` throws an error in TS.

This PR make the options object optional.


erikras:
Published in [`v6.0.1`](https://github.com/final-form/react-final-form/releases/tag/v6.0.1).

PR #503, opened 2019-05-28
Fix Flow typings for UseFieldConfig and FieldProps

eugene-stativka:
There was a mismatch between the source code and Flow typings which led to the following error:
```js
<Field
  name={name}
  component={SomeCustomInputComponent}
  /* ^^^ Flow error here:
  Cannot create Field element because in property component:
   • Either function [1] is incompatible with string literal
     input [2].
   • Or function [1] is incompatible with string literal
     select [3].
   • Or function [1] is incompatible with string literal
     textarea [4].
  */
/>
```

### DIFF
--- a/src/types.js.flow
+++ b/src/types.js.flow
@@ -75,7 +75,8 @@ export type UseFieldConfig = {
   afterSubmit?: () => void,
   allowNull?: boolean,
   beforeSubmit?: () => void | false,
-  component?: SupportedInputs,
+  children?: $PropertyType<RenderableProps<*>, 'children'>,
+  component?: $PropertyType<RenderableProps<*>, 'component'>,
   defaultValue?: any,
   format?: (value: any, name: string) => any,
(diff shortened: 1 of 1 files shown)


erikras:
I feel like ignoring `SupportedInputs` is a mistake. Seems like an important shortlist of valid strings.

eugene-stativka:
> I feel like ignoring `SupportedInputs` is a mistake. Seems like an important shortlist of valid strings.

I am not sure I understood where `SupportedInputs` are ignored.
```js
component?: $PropertyType<RenderableProps<*>, 'component'>, 
// ^^^ this is the same as:
component?: React.ComponentType<*> | SupportedInputs
```
Please clarify.

erikras:
Oh, shit. I didn't know you could do that. 👍

erikras:
Published in [`v6.0.1`](https://github.com/final-form/react-final-form/releases/tag/v6.0.1).

PR #506, opened 2019-05-30
Add ReactFinalFormContext to exported types

kiangtengl:
`ReactFinalFormContext` is exposed in index.js of the library but no such type appears in the libraries' type definition.

This makes it possible to import ReactFinalFormContext from `react-final-form` without any errors.


erikras:
I don't think it _should_ be exported. I think the `useForm()` hook should be used to access the form in the context.

PR #513, opened 2019-06-03
Simplify slightly logic around keeping latest value in ref

Andarist:


src/useLatest.js
```diff
@@ -0,0 +1,12 @@
+// @flow
```
Andarist: I would gladly just install mine - https://github.com/Andarist/use-latest but got an impressions that u dont like external deps here 😉

erikras:

PR #515, opened 2019-06-04
Remove context export

erikras:
In `v4`, we were exporting the `ReactFinalFormContext` to enable consumers to gain access to the form instance from inside the `Form`, but now with the `useForm` hook, we no longer need to export the context value.

Closes #506


erikras:

PR #516, opened 2019-06-04
Strongly typed form values for Flow and Typescript

erikras:
Depends on https://github.com/final-form/final-form/pull/223.

### DIFF
--- a/src/getContext.js
+++ b/src/getContext.js
@@ -1,0 +1,14 @@
+// @flow
+import * as React from 'react'
+import type { FormApi, FormValuesShape } from 'final-form'
+
+let instance: React.Context<?FormApi<any>>
+
+export default function getContext<
+  FormValues: FormValuesShape
+>(): React.Context<FormApi<FormValues>> {
+  if (!instance) {
+    instance = React.createContext<?FormApi<FormValues>>()
+  }
+  return ((instance: any): React.Context<FormApi<FormValues>>)
+}
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -13,8 +13,8 @@ import { Omit } from 'ts-essentials';
 
 type SupportedInputs = 'input' | 'select' | 'textarea';
 
-export interface ReactContext {
-  reactFinalForm: FormApi;
+export interface ReactContext<FormValues> {
+  reactFinalForm: FormApi<FormValues>;
 }
 
 export type FieldMetaState = Omit<
@@ -38,15 +38,15 @@ export interface FieldRenderProps<T extends HTMLElement> {
   meta: FieldMetaState;
 }
 
-export interface FormRenderProps extends FormState {
-  form: FormApi;
+export interface FormRenderProps<FormValues> extends FormState<FormValues> {
+  form: FormApi<FormValues>;
   handleSubmit: (
     event?: React.SyntheticEvent<HTMLFormElement>
   ) => Promise<object | undefined> | undefined;
 }
 
-export interface FormSpyRenderProps extends FormState {
-  form: FormApi;
+export interface FormSpyRenderProps<FormValues> extends FormState<FormValues> {
+  form: FormApi<FormValues>;
 }
 
 export interface RenderableProps<T> {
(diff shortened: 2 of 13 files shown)


typescript/index.d.ts
```diff
@@ -88,22 +88,30 @@ export interface FieldProps<T extends HTMLElement>
...
 
-export interface FormSpyProps
-  extends UseFormStateParams,
-    RenderableProps<FormSpyRenderProps> {}
+export interface FormSpyProps<FormValues>
+  extends UseFormStateParams<FormValues>,
+    RenderableProps<FormSpyRenderProps<FormValues>> {}
 
 export const Field: React.FC<FieldProps<any>>;
 export const Form: React.FC<FormProps<object>>;
```
ChadLefort: I was actually about to submit a PR very similar to this. I just implemented final-form at the place I work and love it. I just wish the types were a bit stronger. While this PR is very close to what I had the `Form` component is still just getting the `object` generic type passed to it.

I actually implemented something like this:
```
export const Form: <FormValues = object>(
  props: FormProps<FormValues>
) => React.FunctionComponentElement<FormProps<FormValues>>;
```
This should allow typescript to infer the types based on the `initialValues` props passed to the form or you could set the generic to the form itself like so:

`<Form<{foo: string, bar: number }> initialValues={initialValues} onSubmit={onSubmit}>`

This similar to what is in #426, but I'm not using a class since it isn't one. 

I also did the same to `Field`:
```
export const Field: <FieldValue = any, T extends HTMLElement = HTMLElement>(
  props: FieldProps<FieldValue, T>
) => React.FunctionComponentElement<FieldProps<FieldValue, T>>;
```

This way I could pass the type `FieldType` to `UseFieldConfig` and then the `initalValue` prop and others on `Field` would be strongly typed:

`<Field<string> name="foo" label="Foo" initialValue="bar" component={TextField} required validate={required} />`

I also did the same thing for field arrays. So when you were mapping through the array and needed to get the value off an index, it would have a type.

erikras: That's very interesting.

I had no idea that you could do `<Field<string> name="foo">` in TSX. That's wild.

ChadLefort: Yea, I've used it a couple of times. I think it's been in typescript since version 2.9 or so. Just allowing for a generic like that could allow something like the Form `component` to infer types sometimes too.

I'd be happy to try and implement the solution above after this PR would get merged. I think at least having generic typing for the `Form` component would be very nice instead of it just being set to `object`.

erikras:

debel27:
Hi @erikras,

I think it would have been more appropriate to land this update in the next major version. Upgrading from 6.0.0 to 6.1.0 may break applications (which was my case), even if it doesn't affect any executable code.

I'm super into it though. Better typings greatly improve developer experience, thanks for the update !

src/getContext.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import * as React from 'react'
+import type { FormApi, FormValuesShape } from 'final-form'
+
+let instance: React.Context<?FormApi<any>>
+
+export default function getContext<
```
Andarist: @erikras what's the reason behind this? kinda surprising, is this lazy evaluation needed for anything?

Andarist:
@debel27 How did it break you? Could you prepare a repro case? Maybe the types could be tweaked to be backward-compatible. However, when dealing with types in JS I wouldn't assume that patch/minor can't break them. TS itself evolves and it's impractical to tie major versioning with types.

erikras:
I debated making this a major version bump or not. If it were "support for redux v7" or some popular external library, it would need to be a major version, but because FF and RFF are so tightly coupled, it didn't feel quite so "breaking".

src/getContext.js
erikras: It's not about lazy, it's about typing. I couldn't figure a way to handle the typing.

I suppose I could've just cast the thing...

debel27:
@Andarist, what broke for me :

- `FormRenderProps` became generic and I had to specify the type parameter
- I needed to upgrade the FF peer dependency

Fixing those took me less than a minute (went from 17 Flow errors to none). 

I was just pointing this out in case you didn't think about it. But it apparently went through the thought process so that's fine by me :)

erikras:
Thanks for pushing back. Good to have people providing such feedback. 👍

PR #520, opened 2019-06-11
Allowed swappable final-form APIs

erikras:
[Suggested in a tweet](https://twitter.com/Syynth/status/1138204720974241792)

> @erikras do you have any interest in allowing the React bindings for @finalformjs to receive the FormApi instance as a prop? I'm interested in it for testing purposes but I could imagine it also having value if someone wanted to pass a custom FF compatible API impl.

It took me a minute to understand what was being asked, but when I did, it seemed like a pretty cool idea. Just as Final Form doesn't mind what view framework consumes it, React Final Form shouldn't mind what underlying engine that comports to the Final Form API is being used.

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -42,6 +42,7 @@ function ReactFinalForm<FormValues: FormValuesShape>({
   debug,
   decorators,
   destroyOnUnregister,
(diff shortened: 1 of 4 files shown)


erikras:

PR #525, opened 2019-06-11
Use the same default for FormValues type in all declared types

Andarist:


erikras:

PR #526, opened 2019-06-12
Replace empty object default type with any object

circlingthesun:
By default FormValues is of type `object` which is empty. When accessing a property on the values object, TypeScript raises the following error:

`Property 'myprop' does not exist on type 'object'.ts(2339)`

This PR resolves this issue.


erikras:

PR #522, opened 2019-06-11
Added tsx generic typings for Field, Form, and FormSpy components

ChadLefort:
As mentioned in https://github.com/final-form/react-final-form/pull/516#discussion_r291005795, this should now allow for TSX generics for `Field`, `Form`, and `FormSpy`. This should allow you to pass in a type to either of these components, or in the case of the `Form` component if you pass in a typed `initialValues` prop typescript will infer the type.

For example:
```typescript
const initialValues: { firstName: string } = { firstName: 'chad' };

<Form initialValues={initialValues} onSubmit={onSubmit}>
  {({ values, form }) => (
    // values and form api get the proper types
  )}
</Form>
```

PR #530, opened 2019-06-14
Strongly typed field values

erikras:
Follow-up to #522, which didn't completely get the JSX generics all the way to the `Field` value.


erikras:

PR #531, opened 2019-06-14
Actual return type for the handleSubmit function

debel27:
Very small pull request that updates the docs to reflect the actual Flow types of `handleSubmit`.

I think some explanation may be added, though. If the function is intended to be passed to the `<form>` element, one might wonder what is the point of returning a promise.


erikras:
Great! Thanks!

PR #537, opened 2019-06-17
Fixed destroyOnUnregister bug

### LINKED ISSUE #523: destroy on unregister breaks initial values
bug report

I set initial value and destroyOnUnregister, when I type to first input, value resets, when I type to second input, value also resets.
Steps to reproduce
set intialValues
set destroyOnUnregister
try to type to field.
You can observe that field value resets to empty state. You should have this behavior at all.

it just resets on unregister
### Sandbox Link
https://stackblitz.com/edit/final-form-issue
https://www.loom.com/share/2147185bb138468290bd8d60d52aac7d

erikras:
Fixes #523, ~~**_but only for initial fields_**~~.

~~Because our fields have to register twice on first render, the `destroyOnUnregister` will naturally destroy the value on first render. This PR fixes the bug for fields that are present on first form render, but will not fix it for new ones added.~~

**Update**: Works now for all fields, regardless of when they are added.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4492,9 +4492,9 @@
       }
     },
     "final-form": {
-      "version": "4.15.0",
-      "resolved": "https://registry.npmjs.org/final-form/-/final-form-4.15.0.tgz",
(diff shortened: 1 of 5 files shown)


zmitry:
But if I have dynamic fields this bug still we appear, right?

erikras:
I might have an idea to avoid that. Stay tuned...

erikras:
Published in [`v6.3.0`](https://github.com/final-form/react-final-form/releases/tag/v6.3.0).

PR #545, opened 2019-06-21
Pass through multiple value to custom components

### LINKED ISSUE #544: `multiple` is not passed down to custom components
I'm using the Dropdown from semantic-ui-react. Just like a normal multiselect dropdown it requires the `multiple` prop to enable multiselect mode.

However, the `Field` component passes `multiple` only to `useField()` but not along with the other props to the component. For `<select multiple>` this is handled correctly in `useField()` (`input.multiple = true`), but if `component` is something custom, then this prop is lost.

Would it make sense to always forward `multiple`, assuming that if it's specified the `input` (whatever it is) knows how to handle it? It doesn't seem to make sense to set `multiple` without passing it down, since `useField` *only* takes it into account for `select` components.

erikras:
Fixes #544.

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -378,6 +378,23 @@ describe('Field', () => {
     expect(getByTestId('name').value).toBe('')
   })
 
+  it('should pass multiple through to custom components', () => {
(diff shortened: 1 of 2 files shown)


erikras:

PR #548, opened 2019-06-24
Do not pass undefined type to input

ThiefMaster:
Some custom widgets (such as the SingleDatePicker in react-dates) have very strict prop types, that reject unknown props. So even `type={undefined}` will fail the prop type check.

With this PR `type` is only included in `input`, if it's not undefined.


erikras:
OH! It took me a minute to figure out what this was doing. Nice PR. 👍

Talk about an edge case! 😄

erikras:

Issue #175, opened 2018-03-01
[bug] typescript FieldRenderProps do not include custom props passed to Field


mxdubois:
### Are you submitting a **bug report** or a **feature request**?
bug report

### What is the current behavior?
The typescript defs assume that only `FieldRenderProps` are passed to `component`, `render`, and `children` (https://github.com/final-form/react-final-form/blob/master/src/index.d.ts#L58-L69) when in fact `FieldRenderProps` AND "any non-API props passed into the <Field/> component" are passed to `component`, `render`, and `children`.

This results in type errors when custom props passed to `Field` are required by the `component`:

> Property 'label' is missing in type 'FieldRenderProps & { children?: ReactNode; }'.

```
23:11 Type '{ name: string; component: (props: FieldLayoutProps) => Element; label: string; Control: string; }' is not assignable to type 'IntrinsicAttributes & FieldProps & { children?: ReactNode; }'.
  Type '{ name: string; component: (props: FieldLayoutProps) => Element; label: string; Control: string; }' is not assignable to type 'FieldProps'.
    Types of property 'component' are incompatible.
      Type '(props: FieldLayoutProps) => Element' is not assignable to type 'string | ComponentClass<FieldRenderProps> | StatelessComponent<FieldRenderProps> | undefined'.
        Type '(props: FieldLayoutProps) => Element' is not assignable to type 'StatelessComponent<FieldRenderProps>'.
          Types of parameters 'props' and 'props' are incompatible.
            Type 'FieldRenderProps & { children?: ReactNode; }' is not assignable to type 'FieldLayoutProps'.
              Property 'label' is missing in type 'FieldRenderProps & { children?: ReactNode; }'.
    21 |       <div>
    22 |         <Field
  > 23 |           name="user.name"
       |           ^
    24 |           component={FieldLayout}
    25 |           label="Full Name"
    26 |           Control="input"
```

...assume `FieldLayoutProps` is minimally defined as:
```
export interface FieldLayoutProps extends FieldRenderProps {
  label: string;
  Control: string;
}
```

### What is the expected behavior?
The typescript defs include "any non-API props passed into the <Field/> component" in `FieldRenderProps` so that there is no compile-time type error when `component` requires props passed to `Field`. 

This would match:
1. the docs: https://github.com/final-form/react-final-form#children-props-fieldrenderprops--reactnode--reactnode
2. the runtime behavior: https://github.com/final-form/react-final-form/blob/master/src/Field.js#L161-L172 

[...]

PR #565, opened 2019-07-19
Issue 175: pass any additional props to custom component

Th0rN13:
This PR for resolve [issue #175](https://github.com/final-form/react-final-form/issues/175)

Now u can add any additional props to your custom component. I use custom Select, so i need to pass label, options and anyAdditionalProp.

### How to use

**RenderForm.tsx**
```
<Field<string>
  name="stack"
  component={SelectAdapter}
  label="Tools"
  options={optionsArrayVariable}
  anyAdditionalProp={anyAdditionalProp}
/>
```

**FormAdapters.tsx (contain adapters for custom components)**
```
interface SelectAdapterProps extends HTMLElement {
  label?: string;
  options: Option[];
  anyAdditionalProp: any;
}

export const SelectAdapter = (props: FieldRenderProps<string, SelectAdapterProps>) => {
  const {
    input,
    meta,
    ...rest
  } = props;
  const {
    options,
    label,
    anyAdditionalProp,
  } = input;
  return (
    <Select
      options={options}
      label={label || ''}
      anyAdditionalProp={anyAdditionalProp}
      {...input}
      {...rest}
    />
  );
};
```

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -31,6 +31,7 @@ interface FieldInputProps<FieldValue, T extends HTMLElement> {
   value: FieldValue;
   checked?: boolean;
   multiple?: boolean;
+  [key: string]: any;
 }
 
 interface AnyObject {


po35:
Thanks for this PR but tbh, we need really some more docs around this (<Field<string>...., FieldRenderProps<string, SelectAdapterProps>, ...extends HTMLElement) Even when I follow your example TS sends errors in my code. Just comparing formik and rff and the type system there (IDK if better or worse) feels somehow more accessible. There I just gave my custom field a type paired with an interface for the props and all fine.  Sorry that I don't have more time to put every error TS gave me. But as said said we should start documenting why @Th0rN13 went the way above in the first place. Thx  CCing @erikras

Yankovsky:
@erikras @Th0rN13 same fix is needed for react-final-form-arrays package.

Yankovsky:
@Th0rN13 @erikras I still have the same issue https://codesandbox.io/s/strongly-typed-form-values-with-react-final-form-ewnfh

erikras:

PR #570, opened 2019-07-24
Allow typed render props in TS

erikras:
Allows providing a shape for `FieldRenderProps`.


erikras:

PR #575, opened 2019-07-30
Add RenderableProps to FormRenderProps

Ben-Ho:
as they are actually added to props for createElement call, destructured
but then re-injected in some cases.

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -44,7 +44,7 @@ export interface FieldRenderProps<FieldValue, T extends HTMLElement> {
 }
 
 export interface FormRenderProps<FormValues = AnyObject>
-  extends FormState<FormValues> {
+  extends FormState<FormValues>, RenderableProps<FormRenderProps<FormValues>> {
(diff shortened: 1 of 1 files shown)


Ben-Ho:
I know this is not a big thing and the most users of react-final-form don't need this props but I stumbled upon this and thought this is better improved here than in my project. (because it shouldn't hurt... :-) )

erikras:
Except they aren't passed, right? If you pass a function as a prop to `render`, it won't be called with itself.

Ben-Ho:
That's right. So in the render-prop case the only added prop is children. In component-prop case it's render and children. In my case I was only using the render-prop. So I should check if types actually take some effect/benefit in the component-prop case. If not it should be enough to add only the optional children-prop to the type-definition.

I choose to add the complete RenderableProps because all props already are defined as optional. But component-prop wont ever be part of this object and therefore shouldn't be even defined as optional in this type. I'll change this and check the component-prop case.

erikras:

PR #578, opened 2019-08-06
pass tests in ReactFinalForm.test.js

threepointone:
DO NOT MERGE without reading.

This PR updates react to the new rc, and uses the new async act to silence a warning in ReactFinalForm.test.js

Further, regarding Field.test.js, I rewrote `sleep` with `act`, and discovered some issues, and disabled some tests -
'should use isEqual to calculate dirty/pristine': This test goes into an infinite loop; it appears act() has caught an actual bug here.
'should `formatOnBlur` most updated value' and 'should pass multiple through to custom components' appear to be breaking some assumptions for react-dom, and should probably be fixed.

I've included the warning messages that show for these warnings too.


threepointone:
Of note, this also updates `testing-library/@react` to include this fix https://github.com/testing-library/react-testing-library/pull/407

Andarist:
Interesting, thanks for this! I'll try to look into those failing test to fix them.

I've looked into `act` implementation now and it seems that it tries to wait (before resolving itself) until current microtask queue finishes and stuff like that. Could you confirm that this is how it makes the whole thing here possible without `act` actually wrapping the `setState` callsite?

threepointone:
That is correct, yes. But to clarify, generally/in sync mode, you don't have to *directly* wrap setState callsites either, anywhere above it (in the function callstack) should be fine.

Andarist:
Yeah, sure - i meant exactly that by wrapping setState. Stack introspection is normally not possible in strict mode after all.

Thanks for confirming that, cant wait for async act now 😄

erikras:
Sweet, some love from the React team! Thanks, @Andarist and @threepointone. 👍

Andarist:
16.9 got released - gonna work on finishing this one soon(-ish). Thanks Sunil for showing how things should be done :)

erikras:
Implemented by #581.

PR #580, opened 2019-08-09
Update Wizard example

TylerRick:
Update Wizard example to use functional Wizard component with hooks instead of a class component


erikras:
The examples are moving into Github, precisely to allow this sort of PR. Can you resubmit one editing the wizard example [here](https://github.com/final-form/react-final-form/tree/master/examples/wizard)?

PR #581, opened 2019-08-09
Fix async act warning

Andarist:
followup to #578


Andarist:
I've noticed that you have broken my PR by merging master to it, so I've fixed it by merging master to it

erikras:

PR #589, opened 2019-08-13
removes Omit from the typescript definitions

keithbloom:
The current typings depend upon `Omit<>` from ts-essentials but this is a common type that has been added by people (myself included) to their projects. The duplicate definition of `Omit<>` causes the compiler to complain. `Omit<>` has also been [added to TS 3.5](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type) so this will cause a conflict for those people who are targeting 3.5

To work around this I have changed the use of the` Omit<>` type helper and replaced it with the more explicit use of `Pick<>` and `Exclude<>`. This will prevent anyone that has their own Omit<> type or is targeting TS 3.5 from failing to build the project. 

The added bonus is that ts-essentials is no longer needed so there is one less dependency required :smiley_cat:. However I decided not to remove this from the `package.json` as the reference to `final-form` is set as to use a `yalc` link. I do not have this setup so decided it is best left out of this PR.

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -9,7 +9,6 @@ import {
   FieldSubscription,
   FieldValidator
 } from 'final-form';
-import { Omit } from 'ts-essentials';
 
 type SupportedInputs = 'input' | 'select' | 'textarea';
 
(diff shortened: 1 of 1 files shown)


erikras:
@Andarist, this look 👍 to you?

Andarist:
Yeah, sure - in the future you could replace it back with builtin Omit. Not sure if it's a great time right now to do it, but maybe once TS@3.6 gets out?

keithbloom:
Thank you @erikras and @Andarist for the review and :+1: .  Can this be now be merged?

erikras:

PR #593, opened 2019-08-15
Migrate one example

CodeWithOz:
- Migrate one example to the `examples` folder.

I only did one so I can be sure that this is the format you want before I do the others. Let me know if anything needs to change. If not, I'll do all the others similarly and send it as one PR.


erikras:
Nailed it! 👍

PR #595, opened 2019-08-17
RFC: pausing validation before unmount

### LINKED ISSUE #408: Validations run on unmount creating unnecessary requests
Bug

When leaving a form that is embedded in a dialog, the form is unmounted which in turn calls unsubscribe and ultimately that runs field validations.  In the case a user cancels on a complex form with server side validations, it creates a noticeable slowdown of ui as well as unnecessary requests to the server.

dashboard_-_archetype_com

Unmount will avoid unnecessary work.

### Sandbox Link
Reproduction based on the `Synchronous field-level validation` https://codesandbox.io/s/pw19j1x52m
1. allow to render initially
2. clear console
3. click  `Unmount form`
4. observe validations firing by added console logs

### What's your environment?
Sandbox updated to latest.
ff: 4.11.0
rff: 4.0.2

### Other information

possibly related #336

asazernik:
I know that with classes this should go in `componentWillUnmount()`, but I'm not 100% clear on the semantics of `useEffect()`. Does it also run parent-first when a tree is being unmounted?

If this does what I expect it to, will fix #408. Will make a test case before removing the RFC tag.

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -106,7 +106,9 @@ function ReactFinalForm<FormValues: FormValuesShape>({
     ]
 
     return () => {
+      form.pauseValidation() // pause validation so we don't revalidate on every field deregistration
       unsubscriptions.forEach(unsubscribe => unsubscribe())
+      // don't need to resume validation here; either unmounting, or will re-run this hook with new deps
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [decorators])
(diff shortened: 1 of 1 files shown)


erikras:
Fantastic! 👍

asazernik:
Heh, alright. I hope you know what the useEffect semantics are better than I do 😅

shrugs:
I can confirm that effect disposers are run parent-first, at least experientially.

erikras:

PR #596, opened 2019-08-19
Lazy evaluation of form and field state

erikras:
Has the added benefit of making all the state values readonly.

That should only be a breaking change for the naughtiest of users.


Andarist:
@erikras are u sure this is safe? seems really unsafe to async react

erikras:
Can you explain why? I'm not sure I understand. How is accessing a getter less "sync" than a property? This also makes them readonly.

erikras:

PR #603, opened 2019-08-25
Fix issue with inline isEqual causing an infinite rerender loop

### LINKED ISSUE #517: Infinite Loop when isEqual is updated
### **Bug report** 

When using \<Field>, if isEqual changes, form re-renders, leading to an infinite loop when creating comparator on the fly. eg:

example: 
```
            <Field
              // You can comment isEqual to stop the loop
			  isEqual={() => console.log("infinite power") || false}
              name="lastName"
              component="input"
              type="text"
              placeholder="Last Name"
            />
```

The behavior changed with the 5.0 release (you can see in the codesandbox below that it does not loop below 5.0)

Infinite loop when comparator function is defined on render.

I'm not sure if this is intended or not. In my case, I can define my function before so it's fine but I'd rather still report the change as it leads to a bug in my app and I could not see it in the changelog.

### Sandbox Link

https://codesandbox.io/embed/react-final-form-simple-example-ed7ic

### What's your environment?

Loop need RFF >= 5.0

### Unrelated

Thanks for your hard work on this awesome library 🚀 !

Andarist:
fixes #517

The same thing still happens for defaultValue & initialValue - but to fix those you'd have to decide what should happen when they change (IMHO nothing should happen, only first values should get registered).

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -895,6 +895,38 @@ describe('Field', () => {
     expect(getByTestId('dirty')).toHaveTextContent('Pristine')
   })
 
+  it('should be able to use inline isEqual to calculate dirty/pristine without falling into infinite rerender loop', () => {
+    const { getByTestId } = render(
+      <Form onSubmit={onSubmitMock} initialValues={{ name: 'bob' }}>
+        {() => (
+          <form>
+            <Field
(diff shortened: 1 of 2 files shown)


erikras:

PR #604, opened 2019-08-25
Remove the version

oliviertassinari:
I have updated the link to work with Material-UI v4.


erikras:

PR #619, opened 2019-09-10
Fix typescript typings in FieldProps

bwlt:
According to [the doc](https://final-form.org/docs/react-final-form/types/FieldProps#parse) the parse prop takes the value from the input (which is typed with `FieldValue`)

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -77,12 +77,12 @@ export interface UseFieldConfig<FieldValue> {
   allowNull?: boolean;
   beforeSubmit?: () => void | boolean;
   defaultValue?: FieldValue;
-  format?: (value: FieldValue, name: string) => any;
+  format?: (value: any, name: string) => any;
   formatOnBlur?: boolean;
   initialValue?: FieldValue;
   isEqual?: (a: any, b: any) => boolean;
   multiple?: boolean;
-  parse?: (value: any, name: string) => FieldValue;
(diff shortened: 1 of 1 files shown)


erikras:

br0wn:
Hello. I ran into issue with this PR when I have different type of value in the input than that is stored in the form values. I get following error:

```
Type '(value: moment.Moment | undefined) => string | undefined' is not assignable to type '(value: Moment | undefined, name: string) => Moment | undefined'.
  Type 'string | undefined' is not assignable to type 'Moment | undefined'.
    Type 'string' is not assignable to type 'Moment | undefined'.
```

```
const parseDate = useCallback(
    (value: Moment | undefined) => value ? value.format('YYYY-MM-DD') : undefined,
    []
  );
```

```
interface FormValues {
  ...
  date?: string;
  ...
}
```

My case converts Moment object to string, but it can be applied to the example from documentation as well:

> parsing localized dates into actual javascript Date objects

Does the `FieldValue` typing represent the value of the field stored in the form or the value put into the input?

Juansasa:
This fix broke everything for us.

We do use custom input and the data types used for the input does not necessary mean the form type.

erikras:
I agree, @Juansasa.

erikras:
Reverted in [`v6.3.3`](https://github.com/final-form/react-final-form/releases/tag/v6.3.3).

Issue #618, opened 2019-09-09
Revers decorators unsubscriptions


adascal:
https://github.com/final-form/react-final-form/blob/3add0f36ecc191ee1b1425142ab34b19679fa00e/src/ReactFinalForm.js#L111

Shouldn't `unsubscriptions` be reversed before doing unsubscribe? 
Then we have multiple decorators rewriting `form.submit` seems like it's restored in a wrong way

For example, I'm using `final-form-focus` and my own

erikras:
Published fix in [`v6.3.1`](https://github.com/final-form/react-final-form/releases/tag/v6.3.1).

PR #620, opened 2019-09-10
fix: decorators unsubscribe order

adascal:
fixes: #618


erikras:

PR #622, opened 2019-09-11
useField: Pass in falsy getValidator where possible

asazernik:
final-form skips a bunch of moderately expensive work when
getValidator() for a field is null. When the user has passed in a falsy
"validate" prop, we should take advantage of this by not passing in a
no-op getValidator() function.


erikras:
I don't think this is valid, because it's not going to reregister if the validator value changes. That's the point of the getter callback.

PR #630, opened 2019-09-18
Fix useFormState and useField Flow types

redbmk:
The [exported flow typing in index.js.flow](https://github.com/final-form/react-final-form/blob/3add0f36ecc191ee1b1425142ab34b19679fa00e/src/index.js.flow#L30) doesn't match [what's actually used in `useFormState`](https://github.com/final-form/react-final-form/blob/3add0f36ecc191ee1b1425142ab34b19679fa00e/src/useFormState.js#L12). When using this externally, this causes Flow to think that `useFormState` might return `null` or `undefined`, when that appears not to be the case. It also thinks that the params are required when they're really optional.

**src/index.js.flow**
```
declare export function useFormState<FormValues>(
  params: UseFormStateParams<FormValues>
): ?FormState<FormValues>
```

**src/useFormState.js**
```
function useFormState<FormValues: FormValuesShape>({
  onChange,
  subscription = all
}: UseFormStateParams<FormValues> = {}): FormState<FormValues> {
```

It looks like a similar change was made in #486 to fix the TypeScript typings in the same way.

**typescript/index.d.ts**
```diff
export function useField<T extends HTMLElement>(
   config: UseFieldConfig
 ): FieldRenderProps<T>;
 export function useForm(): FormApi;
-export function useFormState(params: UseFormStateParams): FormState | void;
+export function useFormState(params?: UseFormStateParams): FormState;
 export const version: string;
```

I'm using this locally with `npm link` and it's working for me. I would add `// @flow` to `useFormState.test.js` and then import from `./index` instead of `./useFormState`, but it seems flow and jest aren't working together out of the box.

### DIFF
--- a/flow-typed/npm/jest_v24.x.x.js
+++ b/flow-typed/npm/jest_v24.x.x.js
@@ -1,0 +1,1202 @@
+// flow-typed signature: ec5f209a3b2ff526e13bf59605e51e87
+// flow-typed version: 6cb9e99836/jest_v24.x.x/flow_>=v0.39.x <=v0.103.x
+
+type JestMockFn<TArguments: $ReadOnlyArray<*>, TReturn> = {
+  (...args: TArguments): TReturn,
+  /**
+   * An object for introspecting mock calls
+   */
+  mock: {
+    /**
+     * An array that represents all calls that have been made into this mock
+     * function. Each call is represented by an array of arguments that were
+     * passed during the call.
+     */
+    calls: Array<TArguments>,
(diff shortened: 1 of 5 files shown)


redbmk:
@erikras I also found that there were issues with the `useField` typings and figured out how to get flow working with jest, so now the tests have flow coverage as well. Since most people will be importing from `index`, I also changed the tests to import from index, which will pull in the exported flow typings.

erikras:

Issue #533, opened 2019-06-14
initialValue of useField() hook resets input.value of the field


ndrsllwngr:
### Are you submitting a **bug report** or a **feature request**?

bug report

### What is the current behavior?

```js
const Component = () => {
  useField("firstName", { subscription: { value: true }, initialValue: "Max" });
  return (
    <div>
      <label>First Name (new)</label>
      <Field
        name="firstName"
        component="input"
        type="text"
        placeholder="First Name"
      />
    </div>
  );
};
```
When changing the input value of `firstName` Field to e.g. 'Max abc' and afterwards replacing this  `Component` with an identical `ComponentCopy` (same fieldName and both using the useField() hook) the changed input.value gets reset to the `initialValue: "Max"` and therefore looses its changes.

![Kapture 2019-06-14 at 16 20 41](https://user-images.githubusercontent.com/10077258/59516099-dfbbc280-8ec0-11e9-8ba8-4908260e69d2.gif)

### What is the expected behavior?

Expected to work like `<Form initialValues={{ firstName: "Max" }} ...>` (which I added to the Sandbox demo at the bottom). When changing the input value of `firstName` Field to e.g. 'Max abc' and afterwards replacing this `Component` with an identical `ComponentCopy` (same fieldName, both use useField() hook to supply initialValue) the changed input.value (which is still part of the form) gets preserved and `ComponentCopy` displays the correct input.value 'Max abc' instead of the `initialValue: "Max"`.

![Kapture 2019-06-14 at 16 18 30](https://user-images.githubusercontent.com/10077258/59515913-7936a480-8ec0-11e9-9954-181ddf34db34.gif)

### Sandbox Link

`firstName` demonstrates the bug and `lastName` presents a hacky solution to get the expected behavior.
At the bottom a standalone form demonstrates the expected behaviour by supplying the `initialValues` to the `<Form ..` directly which is at the moment not achievable by using the `useField()` hook.
https://codesandbox.io/s/react-final-form-simple-example-dmh3h?fontsize=14

### What's your environment?

```json
"final-form": "4.14.1",
"react-final-form": "6.2.0",
"node": ">=10.16.0",
```

ndrsllwngr:
_fyi_ I have cleaned up the Sandbox example, updated the bug report and added improved gifs.

Andarist:
[...]

Issue #536, opened 2019-06-16
Initial Value for a field is set again even if the value for that field has been changed to something else.


vidur149:
### What is the current behavior?
I am using wizrard form. The form in total has three pages. On the basis of field values of the first page, second page field values are generated from an API request. We are sending some default values from our backend and I am setting the initialValue for the field. Everything works fine for the first time but as soon as I navigate to third page and come back to the second page, whatever change I have made to the field values are lost and the fields take up the initial values.

### What is the expected behavior?
The field should retain the changed values.

### 

Second page fields
```
renderCheckboxes = field => (
    <Col className="f-form-row" xs={12} key={field.label}>
      <Row>
        <Col xs={12}><h3>{field.label}</h3></Col>
        {field.options.map(option => (
          <Col xs={12} key={option.value}>
            <Field
              component={CheckboxField}
              name={option.value}
              label={option.label}
              type="checkbox"
            />
          </Col>
        ))}
      </Row>
    </Col>
  )

  renderTextField = field => (
    <Col className="f-form-row" xs={4} key={field.name}>
      <Field
        component={TextField}
        label={field.label}
        name={field.name}
        initialValue={field.default}
        type={field.type}
        required={field.required}
        validate={value => (field.required ? validationhelper.required(value) : undefined)}
      />
    </Col>
  );

  renderSelectField = field => (
    <Col className="f-form-row" xs={4} key={field.name}>
      <Field
        component={SelectField}
        isSearchable
        name={field.name}
        options={field.options}
        initialValue={field.default && {
          label: field.default,
          value: field.default
        }}
        label={field.label}
        required={field.required}
        validate={value => (field.required ? validationhelper.required(value) : undefined)}
      />
    </Col>
  )

  render() {
    const {
      dbConn, setDbConn, platformDetails
    } = this.props;

    if (!platformDetails) {
      return <Loading />;
    }

    const { metadata: { modes, parameters } } = platformDetails;

    return (
      <>
        <Row className="f-form-row">
          <Col className="d-flex align-items-center">
            <h2 className="mdh-light m-0 mr-3">Direct DB Connection</h2>
            <ToggleSwitch
              disabled={modes.length === 0}
              onChange={() => setDbConn(!dbConn)}
              checked={dbConn}
            />
          </Col>
        </Row>
        <Row>
          {parameters.map((field) => {
            if (field.type === 'multiselect' && this.shouldRenderField(field.modes)) {
              return this.renderCheckboxes(field);
            }

[...]

PR #636, opened 2019-09-25
Fix initialValue overrides value if a field is re-mounted

SpadarShut:
This fixes issues #533 and #536.

This PR makes RFF not pass `initialValue` to `registerField` if there is a `value` for the provided key in the form already present.

My use case for this is that my form is rather huge, it's a settings dialog with several tabs, each containing a dozen of fields. Most fields are in separate components and manage their own initial state and onSubmit logic so using initial state on form level and all other logic in the fields is not an option. So when tabs are switched the fields get unmounted and values get reset if user switches back to a tab.

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -59,13 +59,13 @@ function useField<FormValues: FormValuesShape>(
     return beforeSubmit && beforeSubmit()
   })
 
-  const register = (callback: FieldState => void) =>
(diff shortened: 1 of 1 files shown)


erikras:
I think this should be handled in Final Form. https://github.com/final-form/final-form/pull/294

erikras:
Fix published in [`final-form@4.18.6`](https://github.com/final-form/final-form/releases/tag/v4.18.6).

PR #640, opened 2019-10-01
Memoize input and meta

asazernik:
Preventing unnecessary re-renders is the name of the game. Users will only see gains if they pass in a pure/memoized component, but if they do it's a big difference.

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -121,63 +121,61 @@ function useField<FormValues: FormValuesShape>(
     ]
   )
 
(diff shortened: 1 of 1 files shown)


asazernik:
Note on re-checking this - because of extensive whitespace changes, the diffs are best viewed with `--color-words --ignore-space-change`

erikras:
Can you show me a render counting test that passes if `meta` and `input` are memoized but fails if they are not?

PR #649, opened 2019-10-16
Make HTMLElement generic default and optional

ackvf:
Similar to other occurences of default type `T extends HTMLElement = HTMLElement` in the typings file.
It is very handy to not be required to add specific HTMLElement generic if, for example, I am using custom "inputs" or if I only need to use parts of the interface that don't use the second generic.

It is actually more often that I don't need the second generic.


ackvf:
How do I re-run the checks after I have installed the app?

PR #654, opened 2019-10-20
[docs] add new Material-UI mui-rff components

lookfirst:
Thanks for a great library!


lookfirst:
🏓 fyi, a similar change made it into MUI docs... https://github.com/mui-org/material-ui/pull/17943

thanks.

PR #661, opened 2019-10-25
Pass FormValues type to Decorator

sedx:
When from values are typed (suppose `T`) and we pass some decorators, instead `Decorator<T>[]`  decorator prop's type resolves to `Decorator<object>[]`.

So correctly typed decorators with correctly typed form process type error.


erikras:

PR #668, opened 2019-11-04
Remove weird comma in redux-form migration guide

jdeniau:


erikras:
What the heck.. weird...

PR #679, opened 2019-11-19
Revert "Fix typescript typings in FieldProps"

erikras:
Reverts final-form/react-final-form#619


erikras:
Published in [`v6.3.3`](https://github.com/final-form/react-final-form/releases/tag/v6.3.3).

Issue #675, opened 2019-11-15
handleSubmit doesn't return anything


jankalfus:
### Are you submitting a **bug report** or a **feature request**?
Kind-of both I guess.

### What is the current behavior?

Calling `handleSubmit` doesn't return anything. According to the docs it should return a Promise in some cases. What cases, I have no idea. I have tried using an async function, submitting the form with invalid values, valid values, but it still doesn't return anything in any of the cases.

This is really annoying when you want to reset the form after submitting it without errors. I ended up using the following code, but I have no idea if that's the correct approach or not. Not trying to be annoying or anything, but how am I supposed as a user of this library to tell when `handleSubmit` is synchronous and when asynchronous?

```
 <Form<FormValues> onSubmit={onSubmit}>
            {({ handleSubmit, form, valid }): ReactElement => (
                <form
                    className="add-activity"
                    onSubmit={(event): void => {
                        handleSubmit(event);
                        if (valid) {
                            form.reset();
                        }
                    }}
                >

```
### What is the expected behavior?
I'd like to see the behavior of `handleSubmit` standardized. It should either be always asynchronous and return a Promise, or it should be always synchronous and not return anything.

### What's your environment?
react final form 6.3.0, final form 4.18.5

pierreferry:
Hi,
For me (not a TS user), it just 'worked'
`handleSubmit(e).then(reset)`

So this is a typing issue ? 🤔 
Hope this helps a little

jankalfus:
It's not a typing issue, it's unrelated to TypeScript. The problem is that as a user of this library, I have no idea how to tell when handleSubmit returns a Promise (and so runs asynchronously), and when it doesn't (and runs synchronously as in my example). The documentation says nothing on this.

erikras:
Yes, [`handleSubmit` returns `form.submit()`](https://github.com/final-form/react-final-form/blob/master/src/ReactFinalForm.js#L177), which, _if and only if_ your `onSubmit` returns a `Promise`, [does return a `Promise`](https://github.com/final-form/final-form/blob/master/src/FinalForm.js#L1058).

As always, a sandbox demonstrating your problem is worth 1,000 words.

jankalfus:
@erikras Thank you for the explanation! I created a PR to include that to the docs.

marrkeri:
I used to use. Don't forget that onSubmit must be **async**:

```
 const onSubmit = async(values, form) =>{
   await ...
 }

[...]

PR #681, opened 2019-11-21
Added note about the return type of handleSubmit

jankalfus:
This should clarify the handleSubmit's return type to newcomers. Related issue: #675


erikras:

Issue #695, opened 2019-12-12
Submit does not work when <Field /> changes name and have `formatOnBlur`


gbiryukov:
### Are you submitting a bug report or a feature request?
Bug report

### What is the current behavior?
Form submit throws error `Cannot read property 'value' of undefined` when form have field with dynamically changed name and specified `formatOnBlur={true}`

Steps to reproduce:
1. open [sandbox link](https://codesandbox.io/s/usefield-obsolete-context-qvvuy)
2. click "use phone"
3. fill anything into input
4. click submit

### What is the expected behavior?
Form submit just works

### Sandbox Link
https://codesandbox.io/s/usefield-obsolete-context-qvvuy

### What's your environment?
react-final-form@6.3.3
final-form@4.18.6
react@16.12.0

erikras:
Published fix in [`v6.3.4`](https://github.com/final-form/react-final-form/releases/tag/v6.3.4).

PR #697, opened 2019-12-12
Submit edge case fix

gbiryukov:
see #695


erikras:
Wow, what a very edgy case this is! 👍

I've changed your test to use the previously existing `<Toggle/>` component built for exactly this sort of test. 😄

erikras:
Great PR, btw.

- Test that demonstrates the bug ✅
- Fix with lengthy comment explaining ✅

❤️

erikras:

PR #700, opened 2019-12-13
Add missing `to` in philosophy.md

spawnia:


erikras:
To be or not to be... 🤦‍♂️

PR #708, opened 2019-12-27
Improved type definition for Field

jankalfus:
Use Field's first generic argument in the default for second generic argument.

Here's an example (notice any vs Date):
obrazek

Since the first generic argument already defaults to any, there's no need to set any in it's place in the second generic argument. This way, if a developer only sets the first generic argument of Field, they still get the input.value prop properly typed, instead of being any (current behavior). And if they omit it all together, it'll default to any, as it currently does.


erikras:

PR #710, opened 2019-12-30
Fix doc links to FieldRenderProps

nacmartin:
👋These two links are broken in https://final-form.org/docs/react-final-form/types/FieldProps


erikras:
@nacmartin! I know that guy! Feliz año, amigo!

PR #711, opened 2020-01-07
Add data object to useField/Field

PerfectPixel:
This PR adds the data object to useField and Field. It brings the parameters of react-final-form in line with final-form.

Providing the initial value for `data` is useful, if you rely on `meta.data` in your components or validators and want to prevent the initial render without this data.


erikras:

PR #715, opened 2020-01-10
Fix handleSubmit typings

### LINKED ISSUE #86: Issues with Typescript + React native

Feature request

Type errors with react-native + typescript

No type errors :P

### What's your environment?
I'm using react-final-form with react-native
```
"react-final-form": "^2.1.1",
"final-form": "^3.0.0",
"react": "16.0.0",
"react-native": "0.50.4",
```

### Other information

There are two situations where I have type errors.
#### 1. handleSubmit
The current type for handleSubmit is:
```
handleSubmit: (event: React.SyntheticEvent<HTMLFormElement>) => void
```
How can we deal with this in react-native?
onPress events of touchables have this signature:
```
onPress?: (event: GestureResponderEvent) => void;
```

#### 2. Field extra props
Sometimes we need to pass some extra props to the underlying component of Field
```jsx
            <Field
              name='firstName'
              component={TextInputField}
              style={styles.textInputStyle}
            />
```
In this case, the `style` prop is the extra one.
The type system complains because the props that Field is expecting are:
```
export type FieldProps = {
  allowNull?: boolean
  format?: ((value: any, name: string) => any) | null
  parse?: ((value: any, name: string) => any) | null
  name: string
  subscription?: FieldSubscription
  validate?: (value: any, allValues: object) => any
  value?: any
} & RenderableProps<FieldRenderProps>
```
Maybe we should allow anything here? :\ 👎

forivall:
Fixes #86 

The typescript typings work in my local code - in this case, with a button handler, but it should work in React Native typescript typings as well.

Let me know if the flow typings changes are fine - i did my best 😄 .

### DIFF
--- a/src/ReactFinalForm.js
+++ b/src/ReactFinalForm.js
@@ -13,7 +13,7 @@ import type {
   FormValuesShape,
   Unsubscribe
 } from 'final-form'
-import type { FormProps as Props } from './types'
+import type { FormProps as Props, SubmitEvent } from './types'
 import renderComponent from './renderComponent'
 import useWhenValueChanges from './useWhenValueChanges'
 import useConstant from './useConstant'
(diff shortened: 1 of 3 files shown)


forivall:
* ugh, sorry about the formatting changes, the precommit hook did that. Let me know if you want me to force those out of there.

erikras:

Issue #398, opened 2018-12-18
Property "items" missing in FieldRenderProps


dininski:
### Are you submitting a **bug report** or a **feature request**?
**bug report**

### What is the current behavior?

Property "items" missing in FieldRenderProps. When implementing a type-ahead component and using TypeScript, TS complains that the field `items` is missing in FieldRenderProps. The field component seems to have a `  [otherProp: string]: any` property, so it does not error out with `items` as a prop. However, because it extends `RenderableProps<FieldRenderProps>`, which has the following signature:

```typescript
export interface RenderableProps<FieldRenderProps> {
  children?: ((props: FieldRenderProps) => React.ReactNode) | React.ReactNode
  component?: React.ComponentType<FieldRenderProps> | string
  render?: (props: T) => React.ReactNode
}
```
the build errors out and the TS error needs to be ignored.

### What is the expected behavior?

The `items` field should be present in the `FieldRenderProps`, albeit optional. Or there should be an overload for this specific use case.

### Sandbox Link

Unfortunately the issue does not present itself within codesandbox,io

### What's your environment?

A TypeScript React application, bootstrapped with `create-react-app`. TS version: 3.2.2, react-final-form: 4.0.2, final-form: 4.11.0

### Other information

Exact error:
```
Error:(44, 31) TS2339: Property 'items' does not exist on type 'Readonly<{ children?: ReactNode; }> & Readonly<FieldRenderProps>'.
```

iamandrewluca:
https://github.com/final-form/react-final-form/issues/175

myagoo:
Same issue here

erikras:
Published fix in [`v6.3.4`](https://github.com/final-form/react-final-form/releases/tag/v6.3.4).

PR #716, opened 2020-01-13
Add otherProp to FieldRenderProps interface

bhishp:
I believe that the FieldRenderProps interface is typed incorrectly, it should actually allow `[otherProp: string]: any;`. Trying to access props other than `input` or `meta` will currently result in a Typescript error.

As shown in the docs [here](https://final-form.org/docs/react-final-form/types/FieldProps#children) arbitrary props assigned to the `Field` component are passed through to the renderProp.

```js
<Field name="myField" someArbitraryOtherProp={42}>
  {props => {
    console.log(props.someArbitraryOtherProp) // would print 42
    return <input {...props.input}/>
  }}
</Form>
```

We can see the passing of props in the `renderComponent` module here: https://github.com/final-form/react-final-form/blob/master/src/renderComponent.js#L25

# Related issue

Seems an issue is already raised regarding this:

https://github.com/final-form/react-final-form/issues/398

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -42,6 +42,7 @@ interface AnyObject {
 export interface FieldRenderProps<FieldValue, T extends HTMLElement = HTMLElement> {
   input: FieldInputProps<FieldValue, T>;
   meta: FieldMetaState<FieldValue>;
+  [otherProp: string]: any;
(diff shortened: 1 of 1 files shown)


erikras:

Issue #667, opened 2019-11-03
Performance problem with Field or FieldArray during registration and unregistration fields


marrkeri:
### Are you submitting a **bug report** or a **feature request**?

Bug

### What is the current behavior?

Check the sandbox below and click on Initialize button and then on Reset.
Slow registration and unregistration bigger amount of fields(500 rows with 3 columns = 1500 fields). 
I have used subscription only to 'value' and 'submitting'.
I haven't provided any validation yet but keep in mind that it will be also included for the subscription.
I have tried also Field instead of FieldArray. In the sandbox, it is included as a comment.

### What is the expected behavior?

Better performance. 

### Sandbox Link

https://codesandbox.io/s/youthful-morning-msnpk
https://msnpk.csb.app/

### What's your environment?

    "final-form": "^4.18.5",
    "final-form-arrays": "^3.0.1",
    "react": "16.8.6",
    "react-dom": "16.8.6",
    "react-scripts": "3.0.1",
    "react-final-form": "^6.3.0",
    "react-final-form-arrays": "^3.1.1"

### Other information

```<Form debug={console.log}>``` During initialization produce lot of records
![image](https://user-images.githubusercontent.com/39760343/68091348-dab2dc00-fe7e-11e9-871e-370f3fb67f8d.png)

I really love the final form. I have used redux-form, formik, react hook form and nothing is so powerful than your library. Just this performance makes me crazy. Thank you for help.

nj314:
This issue is a major blocker for my current project. Is anyone able to suggest a workaround? My project has up to ~200 fields on the page, and the unmounting time is critical.

ryank109:
It seems to me the notification to listeners happen per field registration and the field registration happens twice in `useField` hook, once for setting up the initial state, and second time for subscribing to the future changes.

I've tried to to make the registration only happen once and manually initialize the state, which did make a noticeable performance improvements.
https://github.com/ryank109/final-form/commit/0f6b989209cbee4cf7646b98a5d70cbbc39a3f52
https://github.com/ryank109/react-final-form/commit/0871e203a33f54110d46e03211885a721177d7f4

joni3k:
Any updates on this?

marrkeri:
Any updates?

coultonluke:
Also in the example given, when you type into the fields it is noticeably slow. This is a problem that I have also experienced with Formik and redux-form... more than 30 fields and it becomes slow.

erikras:
[...]

Issue #639, opened 2019-09-27
Fields are unregistered when not in the DOM


jessicamann:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

Field are unregistered, causing meta data (dirty/valid/etc) to be lost, when they are removed from the DOM. The meta data is regained once the field is brought back into the DOM.

See the code sandbox attached for how the Button cannot property stay enabled when the collapsible panel has been closed once a field value is adjusted.

Steps to reproduce:
1. Open "Address" panel, then type in any value. Notice that the button is now enabled.
2. Collapse "Address" panel. 
Notice that:
- the button is now disabled even though fields have been modified.
- form is considered pristine even though fields have been modified.
- "dirtyFields/visited/modified/touched" under form state is empty though address was modified
3. Open "Address" panel.
Notice that:
- expected form states comes back
- button is now property enabled

### What is the expected behavior?

Fields remain registered upon removal from the DOM, ensuring other subscribers access to it and the form's values and meta states. 

Removing elements from the DOM when not in view can be to optimize view performance; it is not necessarily an intent to remove the field from the form.

### Sandbox Link

https://codesandbox.io/s/expansion-panel-with-field-zusgx

### What's your environment?

react-final-form: 6.3.0
final form: 4.18.5

nosovsh:
That completely makes sense. One more common use case that can not be implemented now: if you want to show error icon in the header of the collapsed panel to show that there are some fields with errors inside it. But of course, you want to show this icon only if this field was touched, otherwise freshly opened form will be already filled with error icons.
Currently there is no way to know whether a field was already touched when you close a panel (-> field dom destroyed -> field is unregistered -> field meta is lost)

nosovsh:
Even more basic use case.
If your form is split by multiple tabs/collapsible panels/modal windows/whatever that can be destroyed and recreated. 
1. User opens one tab, clicks into a required field -> it is touched -> error is shown. 
2. then he opens another tab
3. Comes back to the first one and doesn't see an error message for the first required field because the field's meta is lost

Mottoweb:
I am also getting the error removed from the form state, while submit error is being persisted.

Dragomir-Ivanov:
@Mottoweb 
The way I solve this is just keep dummy registered all fields, all the time.

```
const Dummy = () => {
useField("fieldA")
useField("fieldB")
....
}
```

```
<Form ...>
<Dummy />
....
</Form>
```

PR #727, opened 2020-01-31
Register only once for useField hook

ryank109:
Hi, I'm trying to address an issue describe here on issue #667 to improve performance on initial registration of the field.

Also there's a final-form PR that goes with this change: https://github.com/final-form/final-form/pull/319

Again, to describe the findings, with `useField` hook, it was using `registerField` to register for the first time to get the initial values synchronously, then unregister to throw away the initial subscription to get the initial state, then register again for subscribing to the subsequent updates to the form. And each time it registered, it fired off notify subscription on all previously registered fields and for the current subscription. So, if you are trying to render 100 fields on the screen, that's 2 registration per field and 10100 or n(n + 1) notifications.

So I wanted to separate the registration and subscription to make sure that registration and initial state is calculated during that registration, and subscribe in a separate call to the field states.  With this change, I've reduced the number of notifications to **0**, and only registering the field once.  Another side effect of this change was the fix for this issue as well: https://github.com/final-form/react-final-form/issues/639.  The change I made **does not unregister** the field when it's unmounted.

But..., since I wanted to make sure that it works with currently supported behaviors, I did introduced `keepFieldStateOnUnmount` property to `useField`/`Field` to make sure that the current logic of removing/unregistering the field on unmount still works.


erikras:
Can you see if this is still a problem after [`v6.4.0`](https://github.com/final-form/react-final-form/releases/tag/v6.4.0)?

ryank109:
@erikras 
Yes, seems like it's still a problem.  I've created a repo to test the performance of initial registering: https://github.com/ryank109/react-final-form-pref

react-final-form@6.4.0 with 3000 fields ~7-8s
with this fix, with 3000 fields ~1s

ryank109:
@erikras Any recommendations?

ra2dev:
@erikras
Do you have any objections against this functionality? I'm asking since i want to finish work with this PR and if possible merge it.

Dragomir-Ivanov:
@ra2dev What is the status of this PR. It will be really useful to me, because I have a form, separated in different tabs?
My current workaround if anyone curious, is to keep registered all Form fields, all the time rendering `null`.

@erikras Since FF is not actively maintained, can to transfer rights to anyone who wants.

ra2dev:
@Dragomir-Ivanov 
Was not able to dedicate time on this PR, resolved issue on my side by creating separate context to track registered fields  that are located on "hidden" tabs (temp solution).
Next month planning to work on blog post related to performance of form libraries, hopefully will be able to play with it and mb provide solution

PR #734, opened 2020-02-12
[docs] remove unmaintained mui integration project

lookfirst:
Now that I feel like MUI-RFF is gaining traction and I'm definitely committed to maintaining it, let's remove the link to the old unmaintained project.


lookfirst:
@erikras Ping

PR #737, opened 2020-02-14
Use correct closing tag for <Field> in FieldProps.md

DeadIine:
I've noticed a wrong closing tag in the docs. This PR should fix this. 😄


erikras:
🤦‍♂️ Thanks!

PR #746, opened 2020-02-21
feat: export context

### LINKED ISSUE #745: export context, cant use hooks in class components nor events
### submitting a **feature request**?

It would be great to consume the context directly.

> You can’t use Hooks inside of a class component,
https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

> Call Hooks from React function components.
> Do not call Hooks in class components.
> Do not call in event handlers.
https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions

No direct access to context, not exported.
And can't use hooks in class components.

direct access to context by exporting it, to support access within class components too.

```js
import { ReactFinalFormContext } from 'react-final-form';
```

AndyOGo:
fixes #745

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 // @flow
 import Form from './ReactFinalForm'
 import FormSpy from './FormSpy'
+export { default as ReactFinalFormContext } from './context';
 export { default as Field } from './Field'
 export { default as Form, version } from './ReactFinalForm'
 export { default as FormSpy } from './FormSpy'


erikras:
Why? `useForm` gives you everything you could need.

AndyOGo:
Because hooks cant be used within class components, context types can.

> You can’t use Hooks inside of a class component,
https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

AndyOGo:
@erikras did you got my feedback?

erikras:
I think I'd still rather not. It's easy enough to inject into props with a HOC.

```jsx
const withForm = Component => props => (
  <FormSpy subscription={{}}>
    {({ form }) => <Component {...props} form={form} />}
  </FormSpy>
)
```

AndyOGo:
You are right, it's easy to write a custom `withForm` HoC.
Same is true for this change, it's a one liner, will not affect bundle size and is conform to React Context API. But I can see if you may decide to replace underlying context with something else in the future, this change would introduce problems.

Anyway, thanks for your great work!

Nevertheless it's may also just personal taste, imho I would prefer to consume the context directly.

erikras:
Yeah, my sense is that the context variable should remain internal to the library.

Thanks for your totally reasonable suggestion, though. ❤️

PR #750, opened 2020-02-28
Add button form attribute example to external submit

tkharuk:
Adds an example of external `<button>` using `form` attribute to submit `<form>`


tkharuk:
Not sure how to update Sandbox or is it automagical?

Issue #450, opened 2019-03-29
Not passing name gives object value


nabati:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

If you pass `undefined` to the `Field` `name` prop, you get an object as a value instead of a string. Would be good if it threw an error or a warning instead.

https://codesandbox.io/s/l7xr7qrk2l?fontsize=14

erikras:
Published fix in [`v6.5.0`](https://github.com/final-form/react-final-form/releases/tag/v6.5.0).

PR #777, opened 2020-04-09
if name prop is passed as undefined to Field component, an error is t…

AlexAndriyanenko:
…hrown #450


erikras:
Published in [`v6.5.0`](https://github.com/final-form/react-final-form/releases/tag/v6.5.0).

Issue #779, opened 2020-04-15
React Native application can't press "Next" nor `Go To` buttons


kopax:
### Are you submitting a **bug report** or a **feature request**?

- [x] Bug

### What is the current behavior?

It is not possible to press `Next` or `Go To` buttons.

![image](https://user-images.githubusercontent.com/1866564/79306345-549ec500-7ef5-11ea-8ae3-f9a78117a845.png)

### What is the expected behavior?

I expect to be able to  use those in react native environment.

### What's your environment?

- react-native on expo SDK37

### Other information

I believe this is related to #483 , in order to press the next button, we **MUST** be able to target the underlying reference, see https://stackoverflow.com/questions/32748718/react-native-how-to-select-the-next-textinput-after-pressing-the-next-keyboar

### Possible fix

- Apparently there is a fix already submitted: https://github.com/final-form/react-final-form/pull/608#issuecomment-525540956

It just needs some approval and a release. It is stuck for almost a year, and every RN user would benefits from it.

PR #608, opened 2019-08-27
Forward refs to input component

TylerRick:
Resolves #141 

Still to do:
- [x] Make the same change to Field in `react-final-form-html5-validation` (final-form/react-final-form-html5-validation#18)
- [x] Add a test
- [x] Update docs
- [ ] Get Flow typings to work in Field.js

Issue #483, opened 2019-05-18
forwardRef


viczam:
Bug report

It seems I can't set a ref anymore on `<Form />` components.

I looked into `<ReactFinalForm />` component and it doesn't seem to forward the ref.

Maybe using `useimperativehandle` to expose the form api would be a solution?

erikras:
`<Form>` doesn't, itself, _render_ any components, so I'm not sure what the ref would be too...?

viczam:
Would be nice to access the `form` api through a ref.

erikras:
Seems like you could do a `formRef.current = form` in your `<Form>`'s render prop (or a `<FormSpy>`)... Not gonna be set on first render, but would be from there on. 🤔

viczam:
It looks a bit hackish and limiting. Besides that, what if I don't have to Form's children?
I can probably submit a PR myself toward the end of the week and maybe present some actual use cases if you're ok with that.

erikras:
How about something like this:

```jsx
const MyForm = () => {
  const form = React.useRef()
  return (
    <Form onSubmit={onSubmit} formRef={form}>
      ...
    </Form>
}
```
Where the form element will be courteous enough to pass you back the `form` instance. About what you're looking for?

viczam:
That's exactly what I was looking for. Except that I think it's enough to use the `ref` property and not have a custom `formRef` property for that.
This can probably be achieved with something like this (in `ReactFinalForm.js`):
```js
const ReactFinalForm = forwardRef(({...}, ref) => {
  const form: FormApi = useConstant(() => {
    const f = createForm(config)
    f.pauseValidation()
    return f
  })

  React.useImperativeHandle(ref, form);
});
```

crobinson42:
I have dozens of components that use this:

```
<Form
        ....
        ref={el => (this.form = el)}
/>
```

And since upgrading to v6 of `react-final-form` I'm getting this Warning everywhere:

`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

crobinson42:
FYI - this is a breaking change that needs to be fixed in a patch version ASAP.

In v5 this works:

```
<Form
  onSubmit={() => this.form.form.reset()}
  ref={el => (this.form = el)}
/>
```

and no longer works in v6 due to the ref not being forwarded by `react-final-form`

ChrisWiles:
[...]

PR #781, opened 2020-04-19
Fix: Readme by adding warning for people needing ref

kopax:
Added massive warning to avoid time lost for people wanting to use that library. This will cause a total reimplementation in our application, and should be clearly stated in the README.md

Related to :

- https://github.com/final-form/react-final-form/issues/483
- https://github.com/final-form/react-final-form/issues/779
- https://github.com/final-form/react-final-form/pull/608

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,20 @@ In the interest of making 🏁 React Final Form the best library it can be, we'd
 
 ---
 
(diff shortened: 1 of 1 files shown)


erikras:
Going to merge #608.

PR #787, opened 2020-05-01
re-initialize onFocus/onBlur handlers if the subscribed field changes

rekotiira:
If you use `useField` (or `<Field>`) and the name of the field that you subscribe to changes during the lifetime of the component, the `onFocus` handler will continue calling the `state.focus()`  function of the previous field.

The reason is that `useField` doesn't specify the `state.focus` as a dependency for `React.useCallback`.

It is a common use-case with forms that you, for example, have a list of items, and you want to be able to toggle between the active item that you are editing, but not render all the items in DOM, rather just the one you're editing. At the moment in these use-cases when switching between the active item, the `Field` does not re-mount and the handlers will not work properly.

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -147,7 +147,7 @@ function useField<FormValues: FormValuesShape>(
         }
       },
       // eslint-disable-next-line react-hooks/exhaustive-deps
-      [state.name, format, formatOnBlur]
+      [state.blur, state.name, format, formatOnBlur]
     ),
     onChange: React.useCallback(
       (event: SyntheticInputEvent<*> | any) => {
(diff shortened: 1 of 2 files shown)


rekotiira:
@erikras seems like rollup has dropped support for node8 since 2.0.0:
https://github.com/rollup/rollup/issues/3320

react-final-form's `.travis.yml` config is setup to run on node 8, 10 & 11, and as you can see the node8 build is now failing.

You can easily check this locally by cloning react-final-form and running `npm install` while on version 8. It'll fail even on master. Seems like the breaking change was introduced in rollup@2.3.0, 2.2.0 still works.

Should the node8 be removed from Travis CI config, or should rollup be pinned to older version?

The codesandbox is also failing to something else, but I do not think it is related to my changes. Any idea?

rekotiira:
@erikras I rebased master and now all the tests pass. Could this be merged? It's really annoying having to hack around this.

erikras:
Changing the name of your field while it's mounted seems ill-advised. One could also force an unsubscribe/subscribe cycle with a `key` prop, but okay.

rekotiira:
> Changing the name of your field while it's mounted seems ill-advised. One could also force an unsubscribe/subscribe cycle with a `key` prop, but okay.

We have a form where you have a list of items you can edit, but only one of the items (the selected one) is displayed at a time. We don't on purpose change the name of the field (the name is coming as a prop from parent), it's just optimization React does when it diffs the VDOMs of previous and next state, it notices that the structure is the same and only some attributes for the elements has changed, so it decides to re-use the existing elements instead of unmounting/re-mounting.

The `key` is exactly how we worked around it to force re-mount, but that also has its own share of issues like if you re-mount an element that is scrollable and not scrolled to top, you lose the scroll position etc.

erikras:

PR #789, opened 2020-05-07
[docs] fix "Custom Validation Engine" example

pomeh:
"Custom Validation Engine" example was broken, the error `mutators.setFieldData is not a function` was thrown when you blur a field.

I forked the codesandbox example and fixed it by adding 2 lines:

```
 import Icon from 'react-fontawesome'
 import OnBlurValidation from './OnBlurValidation'
+import setFieldData from 'final-form-set-field-data'

     <Form
       onSubmit={onSubmit}
+      mutators={{ setFieldData }}
```

[Broken example link](https://codesandbox.io/s/kxxw4l0p9o)

[Fixed example link](https://codesandbox.io/s/ok83v)

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -136,7 +136,7 @@ Demonstrates how to use a `FormSpy` component to listen for value changes and au
 
 Demonstrates how to use a `FormSpy` component to listen for values and active field changes to automatically submit values when fields are blurred.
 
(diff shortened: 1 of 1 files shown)


erikras:
Thanks for this. I've updated the original codesandbox. Really need to move them _all_ into Github so PRs can fix them.

PR #793, opened 2020-05-13
Deduce select type from component prop for select-multiple

erikras:
Previously `<Field component="select" multiple>` would generate an error claiming it needs a `type="select"` prop. This was wrong, as we can just see if `component === 'select'` to know.


erikras:
Published in [`v6.5.0`](https://github.com/final-form/react-final-form/releases/tag/v6.5.0).

PR #802, opened 2020-05-27
Update  auto-save-selective-debounce to handle clearing inputs

jdeniau:
The actual example of `auto-save-selective-debounce` does not work when the input is cleared.

It does only send the "save" event when the input is changed but not empty.
This is due to the fact that final-form does not send empty values in the `values` object.


erikras:

PR #806, opened 2020-06-04
Update Typescript definition to match with Final-Form

Lleios:
I added the type variable "InitialFormValues" like in Final-Form definition


erikras:

Issue #858, opened 2020-09-21
Lost the formatting method call in checked field getter


bin1357:
### Are you submitting a **bug report** or a **feature request**?
**bug report**
Lost the formatting method call in getting the checked field value
https://github.com/final-form/react-final-form/blob/master/src/useField.js#L221
it is impossible to make a checkbox that would switch between two true values. For example 'on' and 'off'

### What is the current behavior?
it is impossible

### What is the expected behavior?
it is possible

### Sandbox Link
https://codesandbox.io/s/keen-oskar-l1vku

bin1357:
https://codesandbox.io/s/final-formreact-final-form-simple-forked-f4j86 fixed

bin1357:
```
<Field
    name="employed"
    parse={(t) => (t ? "on" : "off")}
    format={(f) => (f === "on" ? true : false)}
    component="input"
    type="checkbox"
/>
```

ArturKp:
I'm not familiar with the codebase, so can't say anything about the code changes, but this seems to also fix my odd usecase, which I was expecting to work.

```
parse={t => t ? null : 1}
format={f => f === null}
```

Checked: null
Unchecked: 1

erikras:
Published fix in [`v6.5.2`](https://github.com/final-form/react-final-form/releases/tag/v6.5.2).

PR #859, opened 2020-09-21
Lost the formatting method call in checked field getter

bin1357:
Fix issues
https://github.com/final-form/react-final-form/issues/858


erikras:
Published in [`v6.5.2`](https://github.com/final-form/react-final-form/releases/tag/v6.5.2).

Issue #864, opened 2020-10-16
FormProps interface doesn't support custom props.


matiosfree:
### What is the current behavior?
With TypeScript there is an error, when I app extra props.

`<Form
      component={FormComponent}
      onSubmit={handleSubmit}
      onCancel={handleCancel}
    />`

Property 'onCancel' does not exist on type 'IntrinsicAttributes & FormProps<any, Partial<{ firstName: string; }>>'.

### What is the expected behavior?
There should not be any error.

### What's your environment?

"@types/react": "^16.9.49",
"react-final-form": "^6.5.1",

### Other information
For the FieldRenderProps we have:
`[otherProp: string]: any;`

Seems like we need the same for FormProps

erikras:
Published fix in [`v6.5.2`](https://github.com/final-form/react-final-form/releases/tag/v6.5.2).

PR #867, opened 2020-10-21
Update index.d.ts

matiosfree:
#864 FormProps interface doesn't support custom props.


erikras:
Published in [`v6.5.2`](https://github.com/final-form/react-final-form/releases/tag/v6.5.2).

PR #870, opened 2020-10-22
Remove any type for input.onChange

hyzyla:
There two values that can be passed to the `input.onChange`. First is `ChangeEvent` and second is actual value. It would be great if typescript will catch value with incorrect type passed to the `onChange` function.


hyzyla:
@erikras , why pull request was closed?

Issue #431, opened 2019-02-26
parse and format is not working for initialValues


alirezavalizade:
checkout this sandbox

[![Edit 🏁 React Final Form - Parse and Format](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/lrrnjxk5q7?fontsize=14)

mhatvan:
@erikras is there any reasoning why it is not the default for `parse()` and `format()` to be applied for `initialValues`, but rather on `onChange`? 

It seems like 10+ people need this and I just came across this issue at work with no way to work around it.

I could eventually take a look and provide a PR, if desired.

pulse00:
@erikras any feedback on this issue would be highly appreciated. If this is not intentional, we're happy to provide a PR, but we're currently unsure if this should be supported or not.

alirezavalizade:
for me it's not an issue anymore, because I used custom component then I fixed the issue there, just saw this one and wanted to report :)

mrassili:
@erikras this is still an issue, can you explain the reasoning for not running the initial value by format() ?

erikras:
I'm not sure I understand the problem here.

## Value Flow

![flow](https://user-images.githubusercontent.com/4396759/83890340-f3081380-a74b-11ea-97cd-28424aa82200.png)

## Updated Sandbox

Is this not the desired functionality?

[![Edit 🏁 React Final Form - Parse and Format](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/react-final-form-parse-and-format-ow5pb?fontsize=14&hidenavigation=1&theme=dark)

mrassili:
@erikras thanks for the drawing.
the problem is when I want to change the value manually (like when filling a single field by clicking a button) I do `onChange(value)` but the value is not parsed per the `parse` function I gave to the field. same for initialValue it doesn't go through `format`

erikras:
@mrassili Can you provide a sandbox?

reinrl:
@alirezavalizade I took your original sandbox from above (now forked as https://codesandbox.io/s/react-final-form-parse-and-format-forked-tg5zt), and made one big change at line 56 (also had to change the default value to be a string instead of a number):
```js
<Field
  name="phone"
  component="input"
  type="text"
  // you did not have a format prop here originally??
  format={normalizePhone}
  //////////
  parse={normalizePhone}
  placeholder="(999) 999-9999"
/>
```

[...]

PR #873, opened 2020-11-05
Update useField.js (#431)

reinrl:
Parse will both change the associated input's display, as well as the underlying FF value. Previous to this change, Format would only change the associated input's display.


reinrl:
I was having issues running the unit tests locally, so I kicked this off to try and see what failed with my change - now I know. I will wait for more conversation on #431 before we take a run at this again.

PR #893, opened 2021-02-04
Allow decorators prop to change identity, but not array contents, without running on every render

bobbyrenwick:
I am running into an issue of infinite loops when upgrading from pre-hooks code to hooks based forms.

This allows you to have something like the following where the decorators will change identity on every render, without causing that decorator to be unsubscribed and re-subscribed on every render, which seems like it could be unintentional.

I'm unsure how to write a test for this but would appreciate any guidance!

```tsx
const CustomForm = () => {
    return <Form decorators={[decorator1, decorator2]} />
}
```

Tests are failing for something unrelated? `[test]     Cannot find module '@babel/plugin-transform-react-jsx-source'`


erikras:

PR #896, opened 2021-02-26
Fixed Missing dependency

imdkbj:


imdkbj:
Fixed missing dependency prop-types

PR #931, opened 2021-07-21
fix: stabilize useField public handler identity

### LINKED ISSUE #816: input.onChange has a new reference each time value changes

Bug report

Every time a field value changes, onChange gets a new reference (and maybe other event handlers do as well).

When building a custom field and using `useEffect`, it's frequent to depend on `onChange` to update a value. In my case for a large field, I use `useMemo` to make sure I only rerender the parts that need it, but `onChange` makes sure everything gets updated every time the value changes. Is there another way than `useCallback(input.onChange, [])`?

### Sandbox Link

https://codesandbox.io/s/react-final-form-third-party-components-example-lo4yo?file=/index.js

### What's your environment?

(Latest on every lib, as seen on the sandbox)

### Other information

/

Thanks in advance!

kapral18:
`useField` public handlers like `onFocus`, `onChange`, `onBlur` are changing their identities based on various internal variables and externally available props/(config parts) like `value`, `name`, `parse` etc...

I am suggesting that public handlers should not be subject to identity change at all, so that consumers can rely on those without jumping through hoops with `useRef` as in https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often.

As a consequence after this change consumer code relying on this behavior might break, hence this is a BREAKING CHANGE

Closes #816

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -130,11 +130,56 @@ function useField<FormValues: FormValuesShape>(
     ]
   )
 
+  const _valueRef = React.useRef(_value)
+  React.useEffect(() => {
+    _valueRef.current = _value
+  })
+
+  const stateRef = React.useRef(state)
+  React.useEffect(() => {
+    stateRef.current = state
+  })
+
+  const nameRef = React.useRef(name)
+  React.useEffect(() => {
+    nameRef.current = name
+  })
+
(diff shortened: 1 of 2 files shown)


erikras:

PR #933, opened 2021-07-23
Use custom hook to create callback constants

erikras:
Further abstraction around #931.

### DIFF
--- a/src/useConstantCallback.js
+++ b/src/useConstantCallback.js
@@ -1,0 +1,30 @@
+// @flow
+import * as React from 'react'
+
+/**
+ * Creates a callback, with dependencies, that will be
+ * instance === for the lifetime of the component.
+ */
+export default function useConstantCallback(callback, deps) {
+  // initialize refs on first render
+  const refs = deps.map(React.useRef)
+  // update refs on each additional render
+  deps.forEach((dep, index) => (refs[index].current = dep))
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const constant = React.useRef((...args) => {
+    // This if seems weird, but if args is [], then the
+    // first param will be the refs.map()
+    if (args && args.length) {
+      callback(
+        ...args,
+        refs.map(ref => ref.current)
+      )
+    } else {
+      callback(
+        undefined,
+        refs.map(ref => ref.current)
+      )
+    }
+  }, [])
+  return constant.current
+}
--- a/src/useConstantCallback.js
+++ b/src/useConstantCallback.js
@@ -2,29 +2,15 @@
 import * as React from 'react'
 
 /**
- * Creates a callback, with dependencies, that will be
- * instance === for the lifetime of the component.
+ * Creates a callback, even with closures, that will be
+ * instance === for the lifetime of the component, always
+ * calling the most recent version of the function and its
+ * closures.
  */
-export default function useConstantCallback(callback, deps) {
-  // initialize refs on first render
-  const refs = deps.map(React.useRef)
-  // update refs on each additional render
-  deps.forEach((dep, index) => (refs[index].current = dep))
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const constant = React.useRef((...args) => {
-    // This if seems weird, but if args is [], then the
-    // first param will be the refs.map()
-    if (args && args.length) {
-      callback(
-        ...args,
-        refs.map(ref => ref.current)
-      )
-    } else {
-      callback(
-        undefined,
-        refs.map(ref => ref.current)
-      )
-    }
-  }, [])
-  return constant.current
+export default function useConstantCallback(callback) {
+  const ref = React.useRef(callback)
+  React.useEffect(() => {
+    ref.current = callback
+  })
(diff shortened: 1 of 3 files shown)


src/useField.js
```diff
@@ -271,7 +168,65 @@ function useField<FormValues: FormValuesShape>(
...
+           * trust that our `state` closure has the most recent value. This is a problem
+           * if-and-only-if the library consumer has called `onChange()` immediately
+           * before calling `onBlur()`, but before the field has had a chance to receive
+           * the value update from Final Form.
+           */
+          const fieldState: any = form.getFieldState(state.name)
+          state.change(format(fieldState.value, state.name))
+        }
+      },
+      [form, format, formatOnBlur, state]
```
erikras, line 186: Provided in the same order as line 172.

src/useField.js
```diff
@@ -271,7 +168,65 @@ function useField<FormValues: FormValuesShape>(
...
+          }
+        }
+
+        const value: any =
+          event && event.target
+            ? getValue(event, state.value, _value, isReactNative)
+            : event
+        state.change(parse(value, name))
+      },
+      [component, name, parse, state, type, _value]
```
erikras, line 222: Provided in the same order as line 191.

src/useConstantCallback.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Creates a callback, with dependencies, that will be
+ * instance === for the lifetime of the component.
+ */
+export default function useConstantCallback(callback, deps) {
+  // initialize refs on first render
+  const refs = deps.map(React.useRef)
+  // update refs on each additional render
+  deps.forEach((dep, index) => (refs[index].current = dep))
```
iamandrewluca, line 12: Doesn't `refs` get created at each render, this `forEach` line is not needed 🤔

src/useConstantCallback.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import * as React from 'react'
+
+/**
+ * Creates a callback, with dependencies, that will be
+ * instance === for the lifetime of the component.
+ */
+export default function useConstantCallback(callback, deps) {
+  // initialize refs on first render
+  const refs = deps.map(React.useRef)
```
afzalsayed96, line 10: It might be useful to add this line as a safety net for sanity

```js
if (deps.length !== refs.length) {
  console.warn("Length of `deps` shouldn't change between renders")
}
```

src/useConstantCallback.js
afzalsayed96: No they don't. `useRef` is similar to `useState` -- whatever is passed as it's first param only initialises the state/ref in the very first render.

afzalsayed96: My understanding is that you shouldn't update refs in render function as it would break with Suspense. See Dan's comment on this issue: https://github.com/facebook/react/issues/21191#issuecomment-815316815

iamandrewluca: Understood, even if the `deps.map` is called each time, it will still get a new array but with first deps

src/useConstantCallback.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import * as React from 'react'
+
+/**
+ * Creates a callback, with dependencies, that will be
+ * instance === for the lifetime of the component.
+ */
+export default function useConstantCallback(callback, deps) {
```
mpenney99, line 8: Would something like the following help? It will always return the same instance for === purposes, but will call
the latest function passed in. No need to care about callback dependencies at all:

```
export function useWrappedCallback(callback) {
	const ref = React.useRef(callback);
	React.useEffect(() => { ref.current = callback; });
	return React.useCallback((...args) => ref.current.apply(null, args), []);
}
```

src/useConstantCallback.js
mpenney99: I think he probably means why not a single ref containing the deps array, instead of a ref for each dep. Then you don't need the forEach line, or the length check.

erikras: > I think he probably means why not a single ref containing the deps array, instead of a ref for each dep. Then you don't need the forEach line, or the length check.
I wrote this yesterday, and had this thought upon waking this morning. Why not save the whole array? 👍

src/useConstantCallback.js
erikras: Ooh!! I like this a lot!

kapral18: This is epic. I had to stare at it for some time tho. So basically we always return the same identity which is inline with earlier version. BUT we work with inner function which is passed from render which is up to date on every render, thus no need for deps at all indeed and also this scales perfectly, i.e. over time when new vars are added or removed we have to change NOTHING! If I knew about this smart pattern I'd use this instead in previous PR. So I am all for it.

kapral18: The only thing I'd TRY to maybe improve is the name to provide as much info on what this actually does explicit as possible like `useFixedWrapperCallback` or `useCallbackWithFixedWrapper` or `useFixedProxy`. (cc: @erikras)

erikras: Updated PR with your implementation. MUCH nicer!

erikras:

PR #935, opened 2021-08-15
Improved typing for the ability to convert types when using parse/format

### LINKED ISSUE #929: Wrong TS typing of input.value when using parseValue/formatValue that convert types

A bug report I think

So if I make a simple field that stores values of type A and uses some input that accepts values of type B, I get a weird type mismatch.
For example, I have a field of type `Date` that uses a simple html select as input (please see a sandbox link below). I add `parseValue` that gets `string` from select and returns `Date` to store, and I add `formatValue` which does the opposite.
In this case I would expect the `input.value` of `FieldRenderProps` to be a `string` but typescript infers its type as `Date` (in fact it's a string). So I see a TS error.

I would expect the `input.value` of `FieldRenderProps` to be a `string`.

### Sandbox Link

https://codesandbox.io/s/strongly-typed-form-values-with-react-final-form-forked-nfmmq?file=/src/index.tsx

### What's your environment?

react-final-form: 6.3.5
final-form: 4.20.1
typescript: 4.3.2

### Other information

Maybe I'm doing something wrong. Would appreciate if anyone could help me to figure the things out.

mrfratello:
Sometimes it is required that in `FieldState.value` and `input.value` were of different types. This is possible through the use of the parse/format props in Field or in useField config.

But when using typescript type conversion is not performed. `input.value` is always typed as the value from FieldState

This pull request corrects the current typing.

Closes #929

### DIFF
--- a/src/FormSpy.test.js
+++ b/src/FormSpy.test.js
@@ -341,7 +341,7 @@ describe('FormSpy', () => {
             <FormSpy
               subscription={{ dirty: true, values: true }}
               onChange={spy}
-              render={render}
+              render={renderSpy}
             />
           </form>
         )}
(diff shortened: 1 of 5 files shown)


erikras:

PR #947, opened 2021-09-29
fix: correct input.value type for useField and Field

### LINKED ISSUE #946: [6.5.4] Typescript "useField" and "Field" default types

It's more of a bug in the sense that the recent typing changes made some types become `any` which breaks our linting.

image

image

```tsx
// value is of type 'any'
const { input: { value } } = useField<string>('foo');

<Field<string> name="foo">
	// value is of type 'any'
	{({ input: { value } }) => (
		// ...
	)}
</Field>
```

`value` should be of type `string` in both cases, unless it is overiden by passing the `InputValue` type argument (which is not practical, and only works on `useField`).

### Other information

Fix for `useField` is pretty straightforward. Simply set the default type of `InputValue` to `FieldValue`:

```ts
export function useField<
  FieldValue = any,
  T extends HTMLElement = HTMLElement,
  InputValue = FieldValue // <--- here
>(
  name: string,
  config?: UseFieldConfig<FieldValue, InputValue>
): FieldRenderProps<FieldValue, T, InputValue>;
```

It's trickier for field though, because of the order of the parameters.

This doesn't work:
```ts
export const Field: <
  FieldValue = any,
  RP extends FieldRenderProps<FieldValue, T, InputValue> = FieldRenderProps<
    FieldValue,
    HTMLElement,
    InputValue // <--- error: used before definition
  >,
  T extends HTMLElement = HTMLElement,
  InputValue = FieldValue
>(
  props: FieldProps<FieldValue, RP, T, InputValue>
) => React.ReactElement;
```

We would need to reorder the types, but I assume that has some impacts...:
```ts
export const Field: <
  FieldValue = any,
  T extends HTMLElement = HTMLElement,
  InputValue = FieldValue,
  RP extends FieldRenderProps<FieldValue, T, InputValue> = FieldRenderProps<
    FieldValue,
    HTMLElement,
    InputValue // <--- 'any' becomes 'InputValue'
  >
>(
  props: FieldProps<FieldValue, RP, T, InputValue>
) => React.ReactElement;
```

mrfratello:
Fix typing, when using strong lint rules

Closes #946

### DIFF
--- a/typescript/Field.test.tsx
+++ b/typescript/Field.test.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import * as React from 'react';
-import { Field, FieldRenderProps } from 'react-final-form';
+import * as React from "react";
+import { Field, FieldRenderProps } from "react-final-form";
 
 const NumberInput: React.FC<{ value?: number }> = () => null;
 
@@ -17,7 +17,7 @@ function FormText2({ input }: FieldRenderProps<string, HTMLInputElement>) {
 
 function FieldNumberValue() {
   return (
-    <Field<number, FieldRenderProps<number>> name="numberField">
+    <Field<number> name="numberField">
       {({ input }) => <NumberInput value={input.value} />}
     </Field>
   );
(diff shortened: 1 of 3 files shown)


erikras:
Published in [`v6.5.7`](https://github.com/final-form/react-final-form/releases/tag/v6.5.7).

Issue #765, opened 2020-03-30
[Question] Do we still use ts-essentials?


xiaoyu-tamu:
I don't see any code still use ts-essentials. If not, maybe we can remove it from dependencies.

#589 / PR removed usage of ts-essentials
#682 / can be close

PR #948, opened 2021-10-03
feat: 🔄 use StrictOmit

Beraliv:
Hey 👋

I've seen that `Omit` usage was removed here – https://github.com/final-form/react-final-form/pull/589

I was interested whether you want to use builtin `Omit` from `typescript` or use `StrictOmit` from `ts-essentials` which also checks if the property is in the object (before removing it)

If you prefer `Omit` to not use `ts-essentials`, I will adapt the PR to be able to get rid of `ts-essentials`

Refers to https://github.com/final-form/react-final-form/issues/765

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -9,6 +9,7 @@ import {
   FieldSubscription,
   FieldValidator,
 } from "final-form";
+import { StrictOmit } from "ts-essentials";
 
 type SupportedInputs = "input" | "select" | "textarea";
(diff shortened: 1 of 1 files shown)


gertdreyer:
@Beraliv Please use the built-in typescript Omit as per the recommendation https://github.com/final-form/react-final-form/pull/589#issuecomment-521011710

Beraliv:
I'm happy to do it

PR #952, opened 2021-10-17
Update md file by adding a simple example

EZZAHED:
the example is copied form here https://github.com/final-form/react-final-form-hooks


mrfratello:
This example is not suitable, since the hooks are taken from 'react-final-form-hooks'. It uses another api

PR #954, opened 2021-10-21
Allow use of createForm() instance with <Form/>

### LINKED ISSUE #910: Bug: Cannot use createForm() instance with <Form/> normally
After creating a form with `createForm()`:

```ts
const form = createForm({
    initialValues: formValues,
    onSubmit: submitForm,
}),
```

I cannot use it in a normal way in: `<Form form={form} render={...}>`. Instead TypeScript starts complaining about missing onSubmit, then initialValues. Likes I don't pass it the `form` instance.

jamesdphillips:
Fixes unexpected error when using `<Form />` component with an instance of the form api.

  ```
  return (
    <Form
        form={form}
        render={...}
    />
)
  ```

  Screen Shot 2021-10-20 at 9 48 15 PM

---

Closes #910

---

Signed-off-by: James Phillips <jamesdphillips@gmail.com>

### DIFF
--- a/typescript/ReactFinalForm.test.tsx
+++ b/typescript/ReactFinalForm.test.tsx
@@ -192,7 +192,7 @@ function withFormApiInstance() {
   return <Form form={createForm({ onSubmit: noop })} />;
 }
 
-// with form api instance
+// with form api instance & invalid prop
 function withFormApiInstanceAndSubmit() {
   return (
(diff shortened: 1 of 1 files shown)


YegorMy:
That's a pretty important thing for me at the moment. Can we merge that one in if it looks good?

Issue #903, opened 2021-03-23
Bug: Submitting seems to get stuck as `true`


quicksnap:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

Example repo: https://codesandbox.io/s/dreamy-goodall-cvy9c?file=/index.js

Seems that, if `handleSubmit` returns `Promis<void>` without causing an async tick, submission gets stuck in `submitting: true` state.

### What is the expected behavior?

Shouldn't be stuck in submitting; should briefly be `submitting: true` then switch back to false

### Sandbox Link

https://codesandbox.io/s/dreamy-goodall-cvy9c?file=/index.js

### What's your environment?

Latest final-form/rff

### Other information

Hey Erik! 👋 I'll try to dig in at root cause. I have some free time :)

quicksnap:
Seems like the issue happened between 6.0.1 and 6.1.0

quicksnap:
I believe I was able to write a test that replicates it maybe... but I hit a wall at what was going wrong.

quicksnap:
A workaround is to put some kind of `await sleep(1)` statement in `onSubmit` =\

agusterodin:
Did you find some sort of elegant fix? await sleep(0) does work but can make the form contents shift awkwardly for a split second.

rohit-kumar2-simpplr:
I am also facing a similar issue. 
 
I read following from documentation for react-final-form
 a)  `onSubmit` — Point 3 (returns a promise) [https://final-form.org/docs/react-final-form/types/FormProps#onsubmit](url)
 b) `submitting`-  [https://final-form.org/docs/final-form/types/FormState#submitting](url)
I am trying to understand how `submitting` works in react-final-form.

Here’s an example sandbox I created for the same, link: https://codesandbox.io/s/xenodochial-moon-5699r0?file=/index.js

**Issue**: Clicking on Submit again after first submit, keeps `submitting` as true.

NLSVTN:
So was it ever fixed? Have you found workarounds? I am experiencing the same issue with the newest version.

Issue #875, opened 2020-11-09
`submitting` with subscription stuck in `true`


nfantone:
### Are you submitting a **bug report** or a **feature request**?

Bug report.

### What is the current behavior?

If setting `subscription`, under certain circumstances which seem to be related to the time the submit callback takes to complete, `form.submitting` is not reset. It will stay `true` even after the `onSubmit` async callback resolves.

### What is the expected behavior?

`form.submitting` should always flip back to `false` when the promise-returning or `async` `onSubmit` callback resolves.

### Sandbox Link

- Take any of the examples on the website, like [this one](https://final-form.org/docs/react-final-form/examples/subscriptions).
- Comment the `await sleep(300);` call on L:10.
- Try submitting any values with both forms shown.
- Notice the difference between the top and bottom forms: "Submit" button on second form will remain `disabled` due to `submitting`.
- Re-adding the `await sleep(300)` call makes the behaviour go back to normal.

### What's your environment?

- `react-final-form`: `6.5.2`
- `final-form`: `4.20.1`
- `react`: `17.0.1`
- `nodejs`: `v14.15.0`

Running on macOS Catalina `10.15.7 (19H2)`.

correaricardo:
Same here, does anyone know a solution different than unsubscribe?

PR #972, opened 2022-02-09
fix: submitting persisting true state

Karadjordje:
This issue is connected with this PR https://github.com/final-form/react-final-form/pull/513/files

When onSubmit returns without delay update batches and re-render never happens.

This will fix issue with `subscription` and `submitting` where `submitting` will stay `true` even thought the updated value should be `false` 

Closes [#875](https://github.com/final-form/react-final-form/issues/875), [#903](https://github.com/final-form/react-final-form/issues/903)


gertdreyer:
Also closes final-form/final-form#412

Issue #560, opened 2019-07-11
formatOnBlur overloading format/parse original meaning


Andarist:
This API seems conceptually broken right now. With format & parse we have a nice parity:
- format maps a state value into a displayable value (happens during render)
- parse maps displayable value into a state value (happens in a callback)

formatOnBlur though prevents format from being called during render - which so far makes sense, but it calls format inside a callback. So the parity gets lost but this also makes parse weirdly unusable with this option - because it still also gets called inside a callback, so suddenly parse's and format's role is the same (?), they both write to state.

Not sure if the problem is recognizable from this vague description - if not please tell me so and I'll try explain it better.

erikras:
I get it, and even more so, _feel_ it. `formatOnBlur` is a bit of inelegance tacked on to a nicely balanced system.

To solve this, however, we need to come up with a solution to the very legitimate use case of allowing a user to type `24.3` and then, _on blur_, having the field adjust it to `$24.30`, just to give one use case. Any spreadsheet app will demonstrate that. Ideally the value would be stored as `Number`.

Ooooh, just typing that, an idea occurred to me. What if `formatOnBlur` was `formatWhenInactive` (inactive `==` does not have focus). This would allow the formatting to not interrupt the user's typing, and also keep it in render only. The only caveat would be that if your field is subscribed _only_ to `value` and not to `active`, it might not feel the need to rerender when it is blurred. Thoughts? 🤔

Andarist:
I was also thinking about this a little before filing this issue. And putting aside the exact API which would enable this behavior, let's discuss the implementation stuff.

It seems like we agree that `format` should be render-only. This leads us to a natural conclusion that parse can be considered as "pre-commit" thing, whereby commit I mean putting a value into stored form values.

Having those semantics in place another conclusion I have is that parse shouldn't be run on unformatted value, because more often than not such value can be reliably parsed into a stored one. Consider keeping a Date in a form and allowing a manual edit on the linked field - we can't parse a text input back into a Date (to put it inside a form state) in the middle of editing.

The only idea I have right now to solve this is to store the unformatted value inside a component's state separately - sync it with form state by default, but when working in this "formatOnBlur" mode (which I believe is a highly useful feature) use that component's state value instead of the form value when editing and sync them only when we commit that value into a form (so on the blur and similar). 

This would bring generally unwanted complexity into the implementation, but I don't know how to solve this better.

tsufiev:
[...]

PR #974, opened 2022-02-14
Use parse instead of format when calculating checked

jharding:
This may be an incorrect assumption, but it seems like the getter for `checked` should be using `parse`, not `format` when comparing values. In addition to fixing an issue I was experiencing, the TypeScript definitions suggest this should be the intended behavior as well as `parse` returns `FieldValue` and `state.config` is of the type `FieldValue`.

I added a couple of tests to prevent future regressions – please let me know if there's anything else I should do to get this to a state where you'd be comfortable accepting the patch.

Thanks for the great library!

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -161,14 +161,14 @@ function useField<FormValues: FormValuesShape>(
     get checked() {
       let value = state.value;
       if (type === "checkbox") {
-        value = format(value, name);
+        value = parse(value, name);
         if (_value === undefined) {
           return !!value;
         } else {
           return !!(Array.isArray(value) && ~value.indexOf(_value));
         }
       } else if (type === "radio") {
-        return format(value, name) === _value;
+        return parse(value, name) === _value;
       }
       return undefined;
     },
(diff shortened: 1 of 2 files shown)


src/useField.js
```diff
@@ -161,14 +161,14 @@ function useField<FormValues: FormValuesShape>(
     get checked() {
       let value = state.value;
```
nkappler, line 162: value should probably be `const` as you can call `parse` right here, then you could remove line 164 and change 171 to `return value === _value"

nkappler:
I'm not a maintainer, just interested in this PR as I stumbled upon it due to this bug #560

erikras-richard-agent (commented):
**Review by Richard (Team Lead)**

Fix to use `parse` instead of `format` when calculating the `checked` prop. This PR has merge conflicts and is from 2022 — the logic may still be relevant but would need rebasing. Will evaluate whether the fix applies to current code.

erikras-richard-agent:
@coderabbitai review

erikras-richard-agent:
Fix published in [`v7.0.1`](https://github.com/final-form/react-final-form/releases/tag/v7.0.1)

PR #982, opened 2022-03-30
Add React 18 to peer dependencies

yura3d:
React 18 has been released yesterday.

I didn't test React Final Form with new concurrent features, but according to the docs React 18 behaves like 17 if you don't use them. For this case forms in my apps work fine. So I think we can start by updating the peer dependency to avoid package managers warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   },
   "peerDependencies": {
     "final-form": "^4.20.4",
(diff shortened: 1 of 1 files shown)


erikras:
Published in [`v6.5.9`](https://github.com/final-form/react-final-form/releases/tag/v6.5.9).

PR #996, opened 2022-06-20
Update Styles.js

chervyakovru:
Right now forms RenderCounts move to top right corner of window. Added `position: relative` to form for catching counter in form

Now:
Screenshot 2022-06-20 at 1 22 20 PM

Fixed:
Screenshot 2022-06-20 at 1 23 15 PM


chervyakovru:
Check have failed not because of my change

PR #1006, opened 2022-08-17
Fix departing rerender counters

chervyakovru:
Right now forms RenderCounts move to top right corner of window. Added position: relative to form for catching counter in form
Now:
image

Fixed:
image


chervyakovru:
@erikras can you merge it, please? This simple update has already helped several of my friends

PR #1018, opened 2022-12-19
the form stays in `validating` state in a wizard when field is unregistred

iamdey:
Not yet a contribution sorry, it's only a failing test that should not.

With 2 fields in a «wizard» form, if one has an async validation and the other doesn't, the form stays in `validating` state

This PR adds 2 tests:
- 1st with mixed validation sync & async field
- 2nd with only async validation

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -4,6 +4,7 @@ import "@testing-library/jest-dom/extend-expect";
 import { ErrorBoundary, Toggle, wrapWith } from "./testUtils";
 import Form from "./ReactFinalForm";
 import Field from "./Field";
+import { useFormState } from ".";
 
 const onSubmitMock = (values) => {};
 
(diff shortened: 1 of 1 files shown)


xxleyi:
> Rq: I unsuccessfully tried to reproduce this issue in final-form directly.
> It seems related to [runFieldLevelValidation](https://github.com/final-form/final-form/blob/3af52bf3d841e6f8560443597c41367dca8bca9b/src/FinalForm.js#L317) but for now I got lost in the step-by-step debugger

Yes, when some fields unregister, some fields register, final form looks only runFieldLevelValidation, which only notify on field level. When we field level async validation, it in fact will affect form state validating.

So, below code can fix this problem, but I dont know if it is good for other part
```js
    if (hasAsyncValidations) {
      var afterPromise = function afterPromise() {
        state.formState.validating--;
        // key thing is, when validating is 0, notifyFormListeners
        if (state.formState.validating === 0) {
          notifyFormListeners();
        }
        callback();
      };
      promise.then(function () {
        if (nextAsyncValidationKey > asyncValidationPromiseKey) {
          return;
        }
        processErrors(true);
      }).then(afterPromise, afterPromise);
    }
```

iamdey:
Hey, I've missed some notifications but I recently got issues due to my previous workaround.

I've rebased  this PR and now it doesn't fail anymore:



The strange thing is that, I tried to revert changes from @xxleyi to confirm it was fixing the issue but it works too even when commenting the changes made at https://github.com/final-form/final-form/pull/473/files

You can merge this to keep a regression test or close, as you wish.

Cheers.

PR #1034, opened 2023-05-20
fix(useField): Fix no re-render on initialValue change

jspannu919:
Currently useField doesn't retriggers properly if the initialValue is changed (Added test scenario). 
Added explicit check for initial value change to prevent deepEqual overheads.

### DIFF
--- a/src/useField.js
+++ b/src/useField.js
@@ -112,11 +112,14 @@ function useField<FormValues: FormValuesShape>(
 
   React.useEffect(
     () =>
-      register((state) => {
+      register((newState) => {
         if (firstRender.current) {
           firstRender.current = false;
+          if(state.initial != newState.initial){
+            setState(newState);
+          }
         } else {
-          setState(state);
+          setState(newState);
         }
       }, false),
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/src/useField.test.js
+++ b/src/useField.test.js
@@ -455,4 +455,27 @@ describe("useField", () => {
     expect(spy.mock.calls[3][1]).toBe(spy.mock.calls[2][1]); // onFocus
     expect(spy.mock.calls[3][2]).toBe(spy.mock.calls[2][2]); // onBlur
   });
+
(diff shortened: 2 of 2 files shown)


bslipek:
IMO this is fine, you should not listen for `initialValues` after form initialization. Why you need this?

jspannu919:
@bslipek In data driven forms (Ref: https://github.com/data-driven-forms/react-forms ) for conditionally hiding the fields, the fields are wrapped with a Field component so that they get re-rendered when the value in any of the parent attribute change(Ref: https://github.com/data-driven-forms/react-forms/blob/84c4cb8861d689cb87ba68100fe7d3acfd667655/packages/react-form-renderer/src/form-renderer/render-form.js#L47). So, due to the above issue the conditional visibility doesn't works if we reverse the order of the fields(Ref: https://stackblitz.com/edit/1dkk8s-rd1ejn?file=schema.js).

bslipek:
I got your point. But still think that this is final-form problem. But I am just a observer here :)

erikras-richard-agent:
Thanks for this PR! There's an interesting design question here.

**The issue:** `useField` doesn't re-render when `initialValue` prop changes after mount.

**Community perspective (from comments):**  
@bslipek and @EndMove suggest `initialValue` is intentionally for *initial* values only, not dynamic updates. If you need dynamic values, you should update form state directly rather than changing initialValue.

**Code concerns:**
- Uses `!=` instead of `!==` for comparison
- Logic seems backwards: setState only on initial render if initial changed?
- The test doesn't actually verify the behavior works

**Design question for @erikras:**  
Should `initialValue` be reactive (re-initialize when it changes), or is it intentionally only for initial mount?

- **If reactive:** This needs better implementation
- **If not reactive:** Close this PR and document that `initialValue` is for initialization only

My instinct: The name "initialValue" suggests it's for initialization, not dynamic updates. For dynamic defaults, use form-level `initialValues` or `form.change()`.

What's your intent here? 🤔

Issue #1050, opened 2025-07-11
[Bug v7.0.0] useField returns undefined value on initial render despite initialValues being set


yurii-oliiarnyk-finsight:
### Description

Hi, thanks for your work on the library!

We're encountering a blocking issue when trying to migrate to `react-final-form@7.0.0` to prepare for React 19.

### Problem
When `initialValues` are passed to the `<Form />`, the corresponding fields registered via `useField` initially return value: `undefined`. This was not the case in earlier versions.

It seems that the field state is now derived via `form.getFieldState()` during initialization [here](https://github.com/final-form/react-final-form/blob/104b18ed68d30eac1a508473fbc76c42a96682f5/src/useField.ts#L101) — but this runs before the field is registered [here](https://github.com/final-form/react-final-form/blob/104b18ed68d30eac1a508473fbc76c42a96682f5/src/useField.ts#L150-L164), so the field state is not available at that moment.

### As a result:

Fields render with value: undefined on the first render. Components relying on the initial value (e.g., to initialize local state or trigger effects) fail or misbehave.

### Expect result:

Fields render with defined value on initialize as it was before in v6.5.9

### Sandbox Link
v6.5.9 https://codesandbox.io/p/sandbox/h7lk86
Logs:
<img width="294" height="52" alt="Image" src="https://github.com/user-attachments/assets/8abceb4f-ff2d-4d93-931e-752a789acf6e" />

v7.0.0 https://codesandbox.io/p/sandbox/6cd2fz
Logs:
<img width="294" height="52" alt="Image" src="https://github.com/user-attachments/assets/ec0b0324-d1a5-4593-b70a-5422af680390" />

yurii-oliiarnyk-finsight:
What solution do you use for this?
> * Overrides were set for both package to use React-19

chevvy:
> [@chevvy](https://github.com/chevvy) What solution do you use for this?
> 
> > * Overrides were set for both package to use React-19

in my package.json:
```json
{
  dependencies: {
    "final-form": "^4.20.10",
    "react-final-form": "^6.5.9",
  },
  "overrides": {
    "react-final-form": {
      "react": "19.0.0"
    },
    "final-form": {
      "react": "19.0.0"
    }
  },
}
```

Dron007:
[...]

PR #1053, opened 2025-12-02
Fix codesandbox ci

joshua-burbidge:
I was working on a fix for #1050 and may have a PR soon, but thought I would fix the CI first. It's been failing for the past 2 PRs for a few reasons: https://ci.codesandbox.io/status/final-form/react-final-form/pr/1047

 - Deprecated dependency `iltorb` (transitive dependency of bundle-size) causing the yarn install to fail. replaced bundlesize with size-limit which does the same thing and is more actively maintained
     - Note: I increased the listed sizes, but that's only because the actual sizes are bigger than the previous listed thresholds, and previously bundlesize was not actually comparing against the threshold, because [bundlesize uses `maxSize` as the property name](https://github.com/siddharthkp/bundlesize?tab=readme-ov-file#configuration), not `threshold`
 - Config files were not playing nice with the node 18 used by the ci. needed to change file extensions to mjs so they are recognized as ES modules

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -161,6 +161,21 @@ export default [
     },
   },
 
+  // Configuration for .mjs files (ES modules in Node.js environment)
+  {
+    files: ["**/*.mjs"],
+    languageOptions: {
+      ecmaVersion: "latest",
+      sourceType: "module",
+      globals: {
+        ...nodeGlobals,
+      },
+    },
+    rules: {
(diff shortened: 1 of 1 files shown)


erikras-dinesh-agent:
Closing as CodeSandbox is shutting down their CI service. We'll remove the CI configuration from all repos instead of trying to fix it.

Tracking the removal work in Plane tickets.

PR #1060, opened 2026-02-04
Fix #1050: useField returns Form initialValues on first render

### LINKED ISSUE #1050: [Bug v7.0.0] useField returns undefined value on initial render despite initialValues being set

erikras-richard-agent:
**Issue:** In v7.0.0, useField returns undefined on first render even when Form initialValues are set. This breaks apps trying to upgrade from v6.

**Root cause:** useField's useState initializer falls back to creating initial state when getFieldState() returns undefined, but only checks field-level initialValue, not the Form's initialValues.

**Solution:** In the fallback case, check form.getState().initialValues[name] before using field-level initialValue. Form-level initialValues take precedence.

**Tests:**
- useField returns Form initialValues on first render
- useField uses field initialValue when Form initialValues missing that field
- All existing tests pass (14/14)

This is a clean solution that avoids the side effects of synchronous registration during render (which React forbids).

**Closes** #1050

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for field initial-value handling, including nested and array field paths and render-time verification.

* **Bug Fixes**
  * Fields now prioritize form-level initial values with field-level fallback.
  * Multi-selects initialize to an empty array when no initial value is provided.

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -801,16 +801,18 @@ describe("Field", () => {
     );
     expect(red).toHaveBeenCalled();
     expect(red).toHaveBeenCalledTimes(2);
-    expect(red.mock.calls[0][0].input.checked).toBe(false);
-    expect(red.mock.calls[1][0].input.checked).toBe(true); // Correctly true for "red" radio
+    // After fix #1050, initialValues work on first render
+    expect(red.mock.calls[0][0].input.checked).toBe(true); // Correctly true for "red" from initialValues
+    expect(red.mock.calls[1][0].input.checked).toBe(true);
     expect(green).toHaveBeenCalled();
     expect(green).toHaveBeenCalledTimes(2);
     expect(green.mock.calls[0][0].input.checked).toBe(false);
-    expect(green.mock.calls[1][0].input.checked).toBe(false); // Correctly false for "green" radio
+    expect(green.mock.calls[1][0].input.checked).toBe(false); // Correctly false for "green"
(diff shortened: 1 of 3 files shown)


erikras (approved):

Hamund:
Would be great to get this fix! Is there an ETA for a new release?

erikras-richard-agent:

PR #1074, opened 2026-02-13
Fix #974: Use parse instead of format in checkbox/radio checked logic

erikras-richard-agent:
Fixes #974

## Problem
In `useField.ts`, `getInputChecked()` uses `format(value, name)` to transform the value before comparing to `_value`. But `_value` is in internal format while `format` converts to display format. This causes type mismatches (e.g. comparing string "10" to number 10 fails strict equality).

## Solution
Use `parse` instead of `format` in `getInputChecked()`. This compares the parsed internal value with the raw `_value` attribute, ensuring type consistency.

## Tests
Added tests with numeric radio/checkbox values using format/parse to verify the fix and prevent future regressions.

## Summary by CodeRabbit

* **Tests**
  * Added tests to verify correct selection state for radio and checkbox controls when forms are initialized with prefilled values.

* **Bug Fixes**
  * Fixed value comparison logic so radio and checkbox controls reflect initial values correctly by applying the proper value transformation during initialization.

### DIFF
--- a/src/Field.test.js
+++ b/src/Field.test.js
@@ -1249,4 +1249,92 @@ describe("Field", () => {
     }).toThrow("prop name cannot be undefined in <Field> component");
     console.error = consoleError;
   });
+
+  it("should support using format/parse with radio controls", () => {
+    const format = (value) => value && value.toString();
+    const parse = (value) => value && parseInt(value, 10);
+
+    const { getByTestId } = render(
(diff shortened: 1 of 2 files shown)


erikras (changes requested):

erikras (approved):

erikras-richard-agent:

PR #1076, opened 2026-02-13
fix: Call onChange with initial state in useEffect for FormSpy

### LINKED ISSUE #809: FormSpy - Cannot update a component while rendering a different component

bug report

When using `FormSpy`, the `onChange` is called during rendering, and throws an error if you use `setState`.

```js
<FormSpy
  subscription={{ values: true }}
  onChange={change => {
    // fired during rendering, calling a `useState` setter fails
    setValues(change.values);
  }}
/>
```

```
Warning: Cannot update a component (`App`) while rendering a different component (`FormSpy`). To locate the bad setState() call inside `FormSpy`, follow the stack trace as described in https://fb.me/setstate-in-render
    in FormSpy (at App.js:23)
    in form (at App.js:13)
    in ReactFinalForm (at App.js:10)
    in App (at src/index.js:9)
    in StrictMode (at src/index.js:8)
```

No error.

### Sandbox Link

https://codesandbox.io/s/react-final-form-formspy-hltnv?file=/src/App.js

### What's your environment?

```
    "final-form": "4.20.0",
    "react": "16.13.1",
    "react-dom": "16.13.1",
    "react-final-form": "6.5.0",
```

### Other information

The issue seems to be that `onChange` in the `useFormState` is called in the `useState` init function, which means during rendering:
https://github.com/final-form/react-final-form/blob/464f1c7855e93899630df0ad897c322995601849/src/useFormState.js#L26

This is part of a range of error messages introduced in react 16.13, and related to an existing issue (https://github.com/final-form/react-final-form/issues/751)

A workaround would be to put any `setState` on the `onChange` in a `setTimeout(... ,0)`, but this feels like a hack, and not clear to new users.

erikras-richard-agent:
Fixes #809

Problem:
When using FormSpy with an onChange callback that calls setState in a parent component, React throws a warning: 'Cannot update a component while rendering a different component'

Additionally, the current implementation doesn't call onChange with the initial form state because the shallowEqual check in the subscription callback prevents it (initial state === subscribed state).

Root Cause:
1. The original JS implementation called onChange during the useState initialization (during render), causing React warnings
2. The current TS implementation fixed the warning but broke onChange for initial state due to shallowEqual check

Solution:
Added a separate useEffect (with empty deps) that explicitly calls onChange with the initial state after the first render completes. This ensures:
- onChange is called AFTER render (no React warnings)
- onChange IS called with initial state (expected behavior)
- Subsequent changes still trigger onChange via subscription

Changes:
- src/useFormState.ts:
  - Added new useEffect to call onChange(state) after initial render
  - Runs only once with empty dependency array
  - Uses onChangeRef to always call the latest onChange

Impact:
✅ Fixes 'Cannot update while rendering' warning
✅ onChange called with initial state as expected
✅ FormSpy tests pass
✅ No breaking changes

Originally reported in issue #809, attempted fix in PR #965 (closed).

Co-authored-by: Jinxuan Zhu <zhujinxuan@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Form change handlers now run after render and are deduplicated, preventing premature or duplicate onChange notifications. This ensures initial render doesn't trigger callbacks and subsequent updates emit only meaningful state changes, improving consistency and reliability of form state reporting.

### DIFF
--- a/src/useFormState.ts
+++ b/src/useFormState.ts
@@ -14,20 +14,32 @@ function useFormState<FormValues = Record<string, any>>({
   const onChangeRef = React.useRef(onChange);
   onChangeRef.current = onChange;
 
-  // Initialize state with current form state without callbacks
-  const [state, setState] = React.useState<FormState<FormValues>>(() => {
-    // Get initial state synchronously but without callbacks
-    return form.getState();
-  });
+  // Initialize with current form state WITHOUT triggering callbacks during render.
+  // We intentionally use getState() here so render-prop consumers (e.g. <FormSpy>{...})
+  // can read a fully-populated initial state on first render.
+  const [state, setState] = React.useState<FormState<FormValues>>(() =>
+    form.getState(),
+  );
+
+  // We want `onChange` to be called AFTER render (fixes #809) and only with the
+  // subscription-filtered state.
+  const firstSubscriptionRef = React.useRef(true);
+  const pendingOnChangeRef = React.useRef<FormState<FormValues> | null>(null);
+  const lastOnChangeRef = React.useRef<FormState<FormValues> | null>(null);
 
   React.useEffect(() => {
-    // Subscribe to form state changes after initial render
     const unsubscribe = form.subscribe((newState) => {
+      // Ensure we set state at least once from the subscription, even if equal,
+      // so that `onChange` can be fired from an effect after the first render.
+      const isFirst = firstSubscriptionRef.current;
+      if (isFirst) {
+        firstSubscriptionRef.current = false;
+      }
+
+      pendingOnChangeRef.current = newState;
+
       setState((prevState) => {
-        if (!shallowEqual(newState, prevState)) {
-          if (onChangeRef.current) {
-            onChangeRef.current(newState);
-          }
+        if (isFirst || !shallowEqual(newState, prevState)) {
           return newState;
         }
         return prevState;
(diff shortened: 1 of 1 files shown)


erikras-dinesh-agent:
@coderabbitai review

erikras-dinesh-agent:
Thanks for the review @coderabbitai! 

I initially tried implementing your suggestion to move the onChange call outside the setState updater, but it caused test failures. The issue is that when we track state in a ref (), the ref doesn't update until after the component re-renders, which happens asynchronously after the subscription callback fires.

This creates a race condition where we compare  against an outdated , causing incorrect behavior when the onChange callback itself depends on the parent component's render state (as tested in the 'should not rerender when onChange changes' test).

While calling onChange inside the setState updater is not the cleanest pattern, it works correctly because:
1. onChange is called synchronously when state actually changes
2. The callback always sees the correct, up-to-date state
3. React 18's automatic batching handles this gracefully

I've added a comment explaining this trade-off. Happy to discuss alternative approaches if you see a better solution!

erikras-dinesh-agent:
Pushed an update that fixes the CI failures and addresses the onChange timing concerns without calling onChange during render.

Key changes in :
- Keep initial React state from  so render-prop consumers still see fully-populated initial state on first render.
- Fire  **only after render** and **only for subscription-filtered updates** by buffering the latest subscription state in a ref and emitting it in a  effect.
- Ensure we set state at least once from the subscription (even if equal) so the initial  can fire post-render.

Local:  now passes. CI should be green on this push.

erikras-dinesh-agent:
@coderabbitai review

erikras (approved):

erikras-richard-agent:

PR #1079, opened 2026-02-17
Test: Regression tests for submitting not stuck on Promise<void> onSubmit (#903)

erikras-gilfoyle-agent:
## Summary

Regression tests for https://github.com/final-form/react-final-form/issues/903.

## Investigation

The reported bug: when `onSubmit` returns `Promise<void>` (async function with no awaits), `submitting` gets stuck as `true`.

**Finding:** The bug is not reproducible in the current implementation. The notification cycle correctly fires after the Promise resolves, setting `submitting` back to `false`.

## Tests Added

1. `async () => {}` (no awaits) — confirms `submitting` transitions true → false
2. `() => Promise.resolve()` — same verification

Also see: final-form/final-form#548 (same tests at the core level)

## Summary by CodeRabbit

* **Tests**
  * Added regression tests to verify that the form submission state correctly resets after asynchronous submission completion in multiple scenarios.

### DIFF
--- a/src/ReactFinalForm.test.js
+++ b/src/ReactFinalForm.test.js
@@ -990,4 +990,61 @@ describe("ReactFinalForm", () => {
     fireEvent.focus(getByTestId("password"));
     expect(getByTestId("password").value).toBe("f1nal-f0rm-RULEZ");
   });
+
+  it("should set submitting back to false when onSubmit returns Promise.resolve() immediately (#903)", async () => {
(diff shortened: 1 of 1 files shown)


erikras-richard-agent (approved):
Two clean regression tests for #903. Both variants (async void, Promise.resolve) verify submitting resets to false. codecov failure is just threshold noise on a test-only PR.

CI ✅ (real checks) — Approved.

Issue #914, opened 2021-05-06
Nested field with validator gives 


mrassili:
### Are you submitting a **bug report** or a **feature request**?

could be a bug

### What is the current behavior?

I have a nested field that when I attach a validator gives this React error with this stack trace
If I remove the validators from the nested fields it works 

The code basically looks like this:

```
                                <Field
                                  renderer={({ value, onChange }) => (
                                  …
                                    <Field
                                      validate={composeValidators(required())}
                                    />
                                    …
                                  )}
                                  id={…}
                                  name={…}
                                  validate={composeValidators(required())}
                                />

```

Maybe someone can spot the culprit here

```
Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (react-dom.development.js?61bb:23093)
    at scheduleUpdateOnFiber (react-dom.development.js?61bb:21164)
    at dispatchAction (react-dom.development.js?61bb:15660)
    at eval (react-final-form.es.js?9cfd:198)
    at eval (final-form.es.js?f3ac:326)
    at notifySubscriber (final-form.es.js?f3ac:400)
    at eval (final-form.es.js?f3ac:417)
    at Array.forEach (<anonymous>)
    at notify (final-form.es.js?f3ac:409)
    at notifyFormListeners (final-form.es.js?f3ac:854)
    at notify (final-form.es.js?f3ac:1091)
    at runValidation (final-form.es.js?f3ac:706)
    at Object.registerField (final-form.es.js?f3ac:1119)
    at register (react-final-form.es.js?9cfd:490)
    at eval (react-final-form.es.js?9cfd:545)
    at commitHookEffectListMount (react-dom.development.js?61bb:19731)
    at commitPassiveHookEffects (react-dom.development.js?61bb:19769)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:237)
    at invokeGuardedCallback (react-dom.development.js?61bb:292)
    at flushPassiveEffectsImpl (react-dom.development.js?61bb:22853)
    at unstable_runWithPriority (scheduler.development.js?3069:653)
    at runWithPriority$1 (react-dom.development.js?61bb:11039)
    at flushPassiveEffects (react-dom.development.js?61bb:22820)
[...]

PR #1080, opened 2026-02-17
Test: Regression test for nested Field with validator Maximum update depth (#914)

erikras-gilfoyle-agent:
Regression test for https://github.com/final-form/react-final-form/issues/914. Bug not reproducible in current code — nested Fields with validators no longer cause infinite loop.


erikras-richard-agent (approved):
Clean regression test for #914. Nested Fields with validators rendering without throwing Maximum update depth. Simple and effective.

CI ✅ — Approved.

Issue #850, opened 2020-08-25
Warning: Cannot update a component (`ForwardRef(Field)`) while rendering a different component (`ForwardRef(Field)`)


slowselfip:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

A react warning about "invalid set state in render" is shown when rendering the value of a conditionally rendered field that has an initialValue set.

### What is the expected behavior?

That React does not print warnings in the console.

### Sandbox Link

https://codesandbox.io/s/tender-sea-6fmu3?fontsize=14&hidenavigation=1&theme=dark

### What's your environment?

```
{
  "react": 16.13.1,
  "react-dom": 16.13.1,
  "react-final-form": 6.5.1,
  "react-scripts": 3.4.1, 
  "final-form": 4.20.1
}
```

### Other information

Note that the example is a bit contrived and that the real life scenario causing is a bit more complex.

This issue might be related to https://github.com/final-form/react-final-form/issues/828. That issue is regarding a FormSpy and this is related to a conditionally rendered Field with initialValue prop.

Thanks for taking the time and hope this can help in some way.

pmartiner:
Any news on this?

Warrior109:
I still see the same warning.
I Hope it will be fixed.

jaryway:
+1

sophiabarclay:
any news on this?

kure-:
This is still an issue. Complex validation functions that might not be dependent directly on form value/values but on additional data, would still throw that warning when following the documentation and changing the key of the field.
In a state of React where we can simply memoize the validate function, we don't even need final form to do that for us.

antonvialibri1:
If it's related to validation and you have validation on individual form fields, use record-level validation instead:

https://codesandbox.io/s/github/final-form/react-final-form/tree/master/examples/record-level-validation

lovely0806:
I still see the same warning.
I hope it will be fixed.

gertdreyer:
Please see the workaround at https://github.com/final-form/react-final-form/issues/985#issuecomment-1363973861

erikras-richard-agent:
## Investigation

This warning occurs when a conditionally rendered Field with `initialValue` causes a state update during render. The issue is in `useField.ts` where `form.change()` is called inside `useEffect` when the field doesn't exist in form state yet.

### Current Code (useField.ts ~line 163):
```ts
if (!existingFieldState) {
[...]

PR #1081, opened 2026-02-17
Test: Regression tests for #914 (nested validator loop) and #850 (update during render)

erikras-gilfoyle-agent:
Regression tests for:
- https://github.com/final-form/react-final-form/issues/914 — nested Field with validator causing Maximum update depth (not reproducible in current code)
- https://github.com/final-form/react-final-form/issues/850 — Cannot update component while rendering different component (not reproducible in current code)

### DIFF
--- a/src/ReactFinalForm.test.js
+++ b/src/ReactFinalForm.test.js
@@ -991,3 +991,47 @@ describe("ReactFinalForm", () => {
     expect(getByTestId("password").value).toBe("f1nal-f0rm-RULEZ");
   });
 });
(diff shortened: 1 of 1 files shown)


erikras-richard-agent (approved):
Good regression test for #850 — smart use of console.error spy to catch the 'Cannot update component while rendering' warning.

Minor: PR title mentions #914 but that's covered by PR #1080. Consider updating title to just reference #850. Not blocking.

CI ✅ — Approved.

Issue #988, opened 2022-04-29
Radio button remains dirty after submit because `initialValue` does not update to newly saved value despite `initialValue` prop changing


unitytotheunity:
### Are you submitting a **bug report** or a **feature request**?

Bug Report

### What is the current behavior?

Submitting a form with 2 radio button <Field>s results in the radio button state being considered "dirty" after submit succeeds. The form's `initialValues` shows the old/previous value despite the `initialValue` prop on both radio button <Field>s being changed after the submit completes.

### What is the expected behavior?

Given an initial form state of:

`{ section-one: { radio-button: "A" } }`

Given radio buttons initialized like so:
A) (selected)
B)

And a change by the user to:

A)
B) (selected)

Upon clicking "Submit", B would remain as the selected radio button, and the form's `dirty` state would be `false` because `section-one.radio-button` would not be considered `dirty`. (Currently, it _would_ be considered `dirty` because the `initial[name]` would show as being the "old" value instead of the newly saved one.

### Sandbox Link

(Coming soon, hopefully!)

### What's your environment?

final-form@4.20.6
react-final-form@6.5.8
Windows 10

erikras-richard-agent:
## Need More Information

To investigate this issue, I need clarification on a few points:

**1. How are you updating initial values after submit?**

Are you:
- Calling `form.initialize(newValues)` in the `onSubmit` callback?
- Changing the `initialValues` prop on the `<Form>` component?
- Changing the `initialValue` prop on individual `<Field>` components?

**2. What does your code look like?**

Please provide a minimal code example showing:
- Your Form component with initialValues
- Your Field components with radio buttons
- Your onSubmit handler
- How you update initial values after submit

**3. Sandbox Link**

The issue mentions "(Coming soon, hopefully!)" for the sandbox link, but this was opened in 2022. Can you provide a CodeSandbox demonstrating the issue?

**Context:**

Field-level `initialValue` prop only sets the value when the field first registers. To update the form's initial values after submit, you need to either:
- Call `form.initialize(newValues)` to reinitialize the entire form
- Update the `initialValues` prop on the `<Form>` component (which triggers reinitialization)

Without more details, I cannot determine if this is:
- A misunderstanding of how initialValue works
- A real bug in how dirty state is calculated after reinitialization
- An issue specific to radio buttons

Please provide the requested information so I can investigate further.

erikras-richard-agent:
[...]

PR #1085, opened 2026-04-07
Fix #984: useField returns stale values when sibling updates form in useEffect

### LINKED ISSUE #984: Bug: useField can not get newest value
When I change form value in useEffect without event trigger or network delay, useField can not get the newest value.

Demo link is here: https://codesandbox.io/s/react-final-form-usefield-can-not-get-newest-value-gc3do4?file=/index.js

erikras-richard-agent:
## Problem

When a parent/sibling component's `useEffect` changes a form value, other `useField` hooks see stale values because their subscription hasn't registered yet.

The initial state had no-op `blur`/`change`/`focus` handlers, so calls to `input.onChange()` during the effect phase were silently dropped.

## Solution

Replace no-op handlers with live form-backed handlers that call `form.blur`/`form.change`/`form.focus` directly. This ensures effect-time changes propagate immediately, before the permanent subscription is registered.

## Test

Added regression test in `useField.issue-984.test.js` that verifies sibling `useField` hooks receive updated values when another field changes the form in `useEffect`.

Fixes #984

## Summary by CodeRabbit

* **Bug Fixes**
  * Field handlers now correctly propagate focus, blur, change and formatting so related fields stay synchronized; initial-value changes are now detected and applied without disrupting validation.

* **Tests**
  * Added a test that verifies sibling-field updates propagate and that initial-value changes are applied and reflected during component lifecycle.

* **Chores**
  * Linting configuration updated to enable and enforce React Hooks rules.


erikras-richard-agent:
## CodeRabbit Review Comments Addressed

The following issues flagged by CodeRabbit have been fixed in commit d7ffb05:

### `src/useField.issue-984.test.js`

1. **Added `eslint-disable-next-line react-hooks/exhaustive-deps` comment** inside the `ParentWithEffect` `useEffect` to document the intentionally empty dependency array (the effect runs only on mount; `input.onChange` is stable via `useConstantCallback`).

2. **Replaced `.textContent).toBe()` with `.toHaveTextContent()` matcher** — uses the idiomatic `@testing-library/jest-dom` matcher which provides clearer failure messages.

### `src/useField.ts`

3. **Replaced strict reference equality (`!==` / `===`) with `configRef.current.isEqual` (or `(a, b) => a === b` fallback)** for all three comparisons in the FIX #988 `useEffect`:
   - `prevInitialValueRef.current !== initialValue` → `!isEqual(prevInitialValueRef.current, initialValue)`
   - `initialValue !== currentFormInitial` → `!isEqual(initialValue, currentFormInitial)`
   - `currentValue === initialValue` → `isEqual(currentValue, initialValue)`
   
   This ensures custom equality functions (e.g. for object/array `initialValue` props) are respected consistently with how Final Form handles equality elsewhere.

All 153 tests continue to pass after these changes.

erikras-richard-agent:
@coderabbitai review

erikras-richard-agent:
@coderabbitai review

erikras (dismissed):

PR #1086, opened 2026-05-05
ci: add npm publish workflow on GitHub release

erikras-richard-agent:
Adds a GitHub Actions workflow that automatically publishes to npm whenever a GitHub Release is published.

## What it does
- Triggers on  (not on every tag push — requires an explicit GitHub Release to be published, giving you a manual checkpoint before anything hits npm)
- Runs tests and build first, then 
- Uses the  secret for authentication

## Prerequisites
An  secret needs to be added to this repo (or at the org level) before the workflow will work. Use a granular npm automation token scoped to publish access for this package only.

## Summary by CodeRabbit

* **Chores**
  * Automated publishing: releases now trigger an automated publish pipeline.
  * Pre-release handling: pre-releases are published with the "next" tag.
  * Validation before publish: tests and a production build run automatically prior to publishing.
  * Environment update: publishing pipeline now uses Node.js 22.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,37 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


erikras-dinesh-agent (dismissed):
LGTM

PR #1094, opened 2026-06-30
Fix #1093: useField read current value on the render

### LINKED ISSUE #1093: `useField` returns an empty string on the first render for a newly added array item
### Description

After upgrading to newer versions of react-final-form/final-form, `useField()` returns an empty string on the first render for a newly added item inside an array value.

The form values already contain the correct object, but the field hook initially returns "" for the new nested field path. On the next render, it returns the correct value.

This is a regression compared to older versions, where `useField('list.*')` returned the actual value from form state immediately on the first render.

### Steps to reproduce
https://codesandbox.io/p/sandbox/awesome-jepsen-dh35dp

- click add
- a new item is added to the array, item is mounted for list.0
- check the console output from useField("list.0")
- see first empty string, then actual value

```
final-form: '4.20.10,
react-final-form: '6.5.9',
```
Everything ok, the first render returns the value.

Image  

```
final-form: '5.0.1,
react-final-form: '7.0.1',
```
Not ok, the first render is empty.

Image

Maybe related:
https://github.com/final-form/react-final-form/issues/1050

romancavry:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial field state handling so newly added fields (including array paths) display the current value immediately.
  * Corrected dirty/pristine tracking, including more consistent behavior for multi-select fields, so state accurately reflects whether values changed.
* **Tests**
  * Added a test to confirm that newly created array fields render the expected value on first render and that state flags (e.g., dirty/initial) are set correctly.

### DIFF
--- a/src/useField.test.js
+++ b/src/useField.test.js
@@ -599,6 +599,38 @@ describe("useField", () => {
     expect(getByTestId("array").value).toBe("Apple");
   });
 
+  it("should return current form values on first render for newly added array fields", () => {
+    const renderSpy = jest.fn();
+    const MyField = () => {
+      const { input, meta } = useField("items[0].name");
+      renderSpy(input.value, meta.initial, meta.dirty);
+      return <input {...input} data-testid="array" />;
+    };
+    const { getByTestId } = render(
+      <Form onSubmit={onSubmitMock} initialValues={{ items: [] }}>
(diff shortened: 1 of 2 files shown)


romancavry:
Fixes #1094

